PR #87, opened 2022-03-25
Unavailable and New chat pages UI update

iqmulator:


alexis-kochetov (approved):

src/dotnet/Chat.UI.Blazor/Pages/NewChatPage.razor
```diff
@@ -3,22 +3,32 @@
 
 <AuthorizeView Context="auth">
     <Authorized>
-        <div class="px-2 py-2">
-            <div class="mt-5">
-                <EditForm Model="@_model" OnValidSubmit="@OnCreate" >
-                    <DataAnnotationsValidator />
+        <Header>
+            <div class="flex flex-col">
```
crui3er, line 7: I see this piece of code is spreading around.
What if you put to MainLayout and use approach like described in https://stackoverflow.com/questions/64255443/blazor-page-elements-other-than-body

alexyakunin:
Yep, we definitely should do something with this. Originally I was using a similar approach to the one described in SO post, but it didn't work well. Let's talk about this once you have some time.

iqmulator:
Maybe at next meet?

alexyakunin:
The root cause here is that you can't "inject" the component into a "slot" in some other part of the UI - there is no API for this. So we may either come up with such an API or agree we don't have such slots, i.e. we render each subtree independently and there is no way for one subtree to inject some content to another.

alexyakunin:
I think we can come up with a generic RenderArea and RenderIntoArea components talking via some UI service that registers areas & routes renders to the registered ones.

alexyakunin:
Or <Slot Name=...> and <UseSlot Name=...>

PR #89, opened 2022-03-29
Contact list and direct chats

crui3er:


alexis-kochetov:
To mitigate anonimity risks we should ask another person for consent to share contact details. Probably it's not good idea to compose direct chat id from two user ids. Let's discuss.
In general, I expect starting direct conversation should not reveal user identity for authors.

alexis-kochetov (approved):

PR #97, opened 2022-04-05
Feat/direct chat

crui3er:
Peer cahts between chat authors


src/dotnet/Chat.UI.Blazor/Components/ContactLinks.razor
```diff
@@ -18,19 +18,32 @@
 @code {
     public class Model
     {
-        public UserContact[] Contacts { get; set; } = new UserContact[0];
+        public (UserContact, string)[] Contacts { get; set; } = new (UserContact, string)[0];
```
FrolYo, line 21: not clear what the second item (string) is supposed to store

src/dotnet/Chat.UI.Blazor/Components/ContactLinks.razor
```diff
@@ -18,19 +18,32 @@
...
 
     protected override ComputedState<Model>.Options GetStateOptions()
         => new() { InitialValue = new () };
 
     protected override async Task<Model> ComputeState(CancellationToken cancellationToken) {
         var contacts = await Contacts.GetContacts(Session, cancellationToken).ConfigureAwait(false);
+        var list = new List<(UserContact, string)>();
+        foreach (var contact in contacts.OrderBy(c => c.Name)) {
+            string url = contact.Id;
+            var isAuthorsPeerChat = contact.TargetUserId.Value.Contains(":");
```
FrolYo, line 36: not clear why User suddenly becomes a Chat? What I mean is that by user we usually mean very specific entity, don't we?

src/dotnet/Chat.UI.Blazor/Components/ContactLinks.razor
```diff
@@ -18,19 +18,32 @@
...
+            if (isAuthorsPeerChat) {
```
FrolYo, line 37: isn't TryParse meant to check that it's a peer chat along with parsing? like we do it for int or decimal? ```if(int.TryParse....)```

PR #115, opened 2022-04-22
chore: watch mode run for macos

FrolYo:
`dotnet run --project build -- watch` for MacOS

### DIFF
--- a/build/Program.cs
+++ b/build/Program.cs
@@ -86,8 +86,8 @@ internal static class Program
 
         Target("watch", DependsOn("clean-dist"), async () => {
 
-            if (Environment.OSVersion.Platform != PlatformID.Win32NT) {
-                throw new WithoutStackException("Use dotnet watch + webpack watch without build system");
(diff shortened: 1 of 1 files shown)


build/Program.cs
```diff
@@ -86,8 +86,8 @@ internal static class Program
 
         Target("watch", DependsOn("clean-dist"), async () => {
 
-            if (Environment.OSVersion.Platform != PlatformID.Win32NT) {
-                throw new WithoutStackException("Use dotnet watch + webpack watch without build system");
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && RuntimeInformation.IsOSPlatform(OSPlatform.OSX)) {
```
alexis-kochetov, line 89: I don't get it = `if windows _AND_ OSX -> throw exception`.
Originally the exception was thrown when the OS IS NOT WINDOWS

alexis-kochetov (changes requested):

build/Program.cs
FrolYo: my bad. fixed

alexis-kochetov (approved):

PR #118, opened 2022-04-22
fix: default chat for admins restored #117

### LINKED ISSUE #117: fix: admins after logging in are redirected to the-actual-one chat which is not available to them
**Issue:**
Instead of showing default chat at http://localhost:7080/chat/the-actual-one it redirects to http://localhost:7080/unavailable/chat

FrolYo:
Fixes #117 
ShouldRecreateDb was commented (I guess accidentally @alexis-kochetov). That led to ChatDbInitializer skipping, so the default chat was not created on local dev env and that led to redirect to `unavailable`


alexyakunin:
Thx!

alexyakunin (approved):

alexis-kochetov (approved):
It's my fault. But eventually we have to implement checks that will trigger db creation if it doesn't exist

PR #152, opened 2022-05-11
feat: add common escape handler

evgeniy-ac:
#151


src/dotnet/UI.Blazor/Services/EscapeHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+                elementRef,
+                _blazorRef)
+            .ConfigureAwait(true);
+
+    public async ValueTask DisposeAsync()
+    {
+        if (_jsRef == null)
+            return;
+
+        await _jsRef.InvokeVoidAsync("dispose").ConfigureAwait(true);
```
FrolYo, line 36: not sure if it matters for us, but looks like disposing twice leads to null ref inside js

src/dotnet/UI.Blazor/Components/EscapeHandler/escape-handler.ts
```diff
@@ -0,0 +1,26 @@
+export class EscapeHandler {
+    private _blazorRef: DotNet.DotNetObject;
+    private _elementRef: HTMLDivElement;
```
alexyakunin, line 3: Why it's a div element - e.g. not `body` or any element?

src/dotnet/UI.Blazor/Components/Modal/DialogFrame.razor
```diff
@@ -1,19 +1,43 @@
...
 
-<ModalFrame Class="@Class" Title="@Title" HasHeader="HasHeader" HasCloseButton="HasCloseButton">
-    <div class="dialog-body">
-        @Body
-    </div>
-    <div class="dialog-buttons">
-        @Buttons
-    </div>
-</ModalFrame>
+<div @ref="_elementRef" tabindex="-1" autofocus>
```
alexyakunin, line 5: Ok, I guess I get the part on element kind now, but...

src/dotnet/UI.Blazor/Services/EscapeHandler.cs
```diff
@@ -0,0 +1,38 @@
+using ActualChat.UI.Blazor.Module;
+
+namespace ActualChat.UI.Blazor.Services;
+
+public sealed class EscapeHandler : IAsyncDisposable
```
alexyakunin, line 5: So I'd rather turn this into a "headless" scoped service like ModalUI or ClipboardUI, which attaches the handler to HTML `body` element rather than affects rendering in any way.

alexyakunin (changes requested):

src/dotnet/UI.Blazor/Services/EscapeHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    public EscapeHandler(IJSRuntime jsRuntime)
+    {
+        _blazorRef = DotNetObjectReference.Create(this);
+        _jsRuntime = jsRuntime;
+    }
+
+    [JSInvokable]
+    public void OnEscape()
+        => Escape?.Invoke(this, EventArgs.Empty);
```
alexyakunin, line 22: One other downside of this design is that you can't handle escape purely in JS - e.g. you can't hide the content there on escape & don't even trigger any interop (remember it's a networking call in Blazor Server) when you need it. And it might be a frequent scenario - e.g. you can at least hide almost any element instantly first, and only after that trigger the Blazor call to do the rest (i.e. actually remove the content from DOM & dispose stuff).

So that's why I said it makes sense to implement a JS-based API  - I was assuming you start from it & add its Blazor counterpart after.

alexyakunin (changes requested):

src/dotnet/UI.Blazor/Components/Modal/DialogFrame.razor
evgeniy-ac: we need `tabindex="-1"` to be able to focus an element

src/dotnet/UI.Blazor/Services/Escapist.cs
```diff
@@ -0,0 +1,17 @@
...
+public sealed class Escapist
+{
+    private readonly Func<EscapistSubscription> _factory;
+
+    public Escapist(Func<EscapistSubscription> factory)
+        => _factory = factory;
+
+    public async Task<IAsyncDisposable> SubscribeAsync(Action action, CancellationToken token)
+    {
+        if (action == null) throw new ArgumentNullException(nameof(action));
```
alexyakunin, line 12: We rarely do this - i.e. I guess we don't do null checks when the argument is non-nullable + null will still trigger an error later. Sorta "too much of a fool-proof" thing.

alexyakunin (approved):
LGTM with minor changes.

src/dotnet/UI.Blazor/Services/Escapist.cs
evgeniy-ac: done

src/dotnet/UI.Blazor/Components/Escapist/disposable.ts
```diff
@@ -0,0 +1,3 @@
+export interface Disposable {
```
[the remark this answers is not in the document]
evgeniy-ac: done

PR #165, opened 2022-05-12
Feat/remove user authors

crui3er:
- removed UserAuthors/UserAuthorsBackend
- removed DbUserAuthor
- moved AvatarId property from DbUserAuthor to DbUserProfile
- simplify DbUserProfile creation


src/dotnet/Users.Contracts/Backend/IUserProfilesBackend.cs
```diff
@@ -5,18 +5,11 @@ public interface IUserProfilesBackend
     [ComputeMethod(KeepAliveTime = 10)]
     Task<UserProfile?> Get(string id, CancellationToken cancellationToken);
     [ComputeMethod(KeepAliveTime = 10)]
-    Task<UserProfile?> GetByName(string name, CancellationToken cancellationToken);
```
FrolYo, line 8: Let's also remove client defs

FrolYo (approved):

src/dotnet/Users.Contracts/Backend/IUserProfilesBackend.cs
crui3er: Updated IUserProfilesBackendClientDef. But it seems it's not used for now.

PR #178, opened 2022-05-16
fix: markup for new message panel and right settings

evgeniy-ac:
#175


alexyakunin (approved):
I can't see the effect, but it's fine otherwise

PR #181, opened 2022-05-17
chore: audio context lazy

FrolYo:
After some discussion we came to decision that we don't need audio context pool, but instead we need just lazy audio context


src/dotnet/Users.UI.Blazor/Pages/UserInvitePage.razor
```diff
@@ -27,7 +27,7 @@
 
     private async Task OnNewInviteClick() {
         var invite = new Invite {
-            Remaining = 10,
+            Remaining = 50,
```
alexis-kochetov, line 30: Please roll-back unrelated changes

src/nodejs/src/audio-context-lazy.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    private async create() {
+        const audioContext = new AudioContext({
+            latencyHint: 'interactive',
+            sampleRate: 48000,
+        });
+
+        console.log(`${LogScope}.initialize: audioContext.state=`, audioContext.state);
+        await audioContext.resume();
```
alexis-kochetov, line 32: Are you sure you don't need anything like that 
```
    if (audioContext.state === 'suspended')
        await audioContext.resume();
```

src/nodejs/src/audio-context-lazy.ts
```diff
@@ -0,0 +1,82 @@
...
+    constructor() {
+        this.addInitEventListeners();
+    }
+
+    public async get(): Promise<AudioContext> {
+        if (!this.audioContext) {
+            await this.whenInitialized;
+        }
+
+        return this.audioContext;
```
alexis-kochetov, line 16: Probably we need to check state there and call resume if state is not expected

alexis-kochetov (changes requested):

src/dotnet/Users.UI.Blazor/Pages/UserInvitePage.razor
FrolYo: ok...

src/nodejs/src/audio-context-lazy.ts
FrolYo: done

src/nodejs/src/audio-context-lazy.ts
FrolYo: done

alexis-kochetov (approved):

PR #188, opened 2022-05-23
feat: auto turn off recording #174

### LINKED ISSUE #174: feat: auto turn-off recording
2 minutes of silence

FrolYo:
Audio recording is automatically turned off after some inactivity (2 min by default) from all of the chat participants. By inactivity we mean that there is nothing transcribed from recorded audio from any chat member. That means that even if there are still some sounds produced but but they do not contain any recognisable speech the recording will be turned off, since those sounds will be interpreted as a meaningless noise.

Resolves #174

### DIFF
--- a/src/dotnet/Audio.Service/Processing/OpenAudioSegment.cs
+++ b/src/dotnet/Audio.Service/Processing/OpenAudioSegment.cs
@@ -47,14 +47,6 @@ public sealed class OpenAudioSegment
                 Index, StreamId);
     }
 
-    public void SetAudibleDuration(TimeSpan audibleDuration)
-    {
-        if (!TaskSource.For(AudibleDurationTask).TrySetResult(audibleDuration))
-            Log.LogWarning(
-                "SetAudibleDuration came too late for OpenAudioSegment #{Index} of Stream #{StreamId}",
-                Index, StreamId);
-    }
-
     public void Close(TimeSpan duration)
     {
         TaskSource.For(RecordedAtTask).TrySetResult(null);
--- a/src/dotnet/Audio.Service/Processing/OpenAudioSegment.cs
+++ b/src/dotnet/Audio.Service/Processing/OpenAudioSegment.cs
@@ -47,6 +47,15 @@ public sealed class OpenAudioSegment
                 Index, StreamId);
     }
 
+    // TODO(AK): review: use or delete
+    public void SetAudibleDuration(TimeSpan audibleDuration)
+    {
+        if (!TaskSource.For(AudibleDurationTask).TrySetResult(audibleDuration))
+            Log.LogWarning(
+                "SetAudibleDuration came too late for OpenAudioSegment #{Index} of Stream #{StreamId}",
+                Index, StreamId);
+    }
+
     public void Close(TimeSpan duration)
     {
         TaskSource.For(RecordedAtTask).TrySetResult(null);
@@ -58,6 +67,7 @@ public sealed class OpenAudioSegment
         TaskSource.For(ClosedSegmentTask).SetResult(audioSegment);
     }
 
+    // TODO(AK): review: use or delete
     public void TryClose(Exception error)
     {
         TaskSource.For(RecordedAtTask).TrySetException(error);
--- a/src/dotnet/Chat.UI.Blazor/Services/ChatUIStateSync.cs
+++ b/src/dotnet/Chat.UI.Blazor/Services/ChatUIStateSync.cs
@@ -1,4 +1,5 @@
 ﻿using ActualChat.Audio.UI.Blazor.Components;
+using ActualChat.Chat.Module;
 using ActualChat.Users;
 
 namespace ActualChat.Chat.UI.Blazor.Services;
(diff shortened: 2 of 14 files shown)


alexis-kochetov (approved):

src/dotnet/Chat.UI.Blazor/Services/ChatUIStateSync.cs
```diff
@@ -113,18 +120,62 @@ async ValueTask<bool> IsLanguageChanged()
...
+        var recordingChatId = await ChatUI.RecordingChatId.Use(cancellationToken).ConfigureAwait(false);
+        if (recordingChatId.IsEmpty)
+            return (recordingChatId, 0);
+
+        var (_, end) = await Chats.GetIdRange(Session, recordingChatId, ChatEntryType.Text, cancellationToken).ConfigureAwait(false);
+        return (recordingChatId, end);
+    }
+
+    private Task RestartRecording(
+        bool stop,
```
alexyakunin, line 138: mustStop?

src/dotnet/Chat.UI.Blazor/Services/ChatUIStateSync.cs
```diff
@@ -113,18 +120,62 @@ async ValueTask<bool> IsLanguageChanged()
...
+    }
+
+    [ComputeMethod]
+    protected virtual async Task<(Symbol ChatId, long LastEntryId)> GetLastChatEntry(CancellationToken cancellationToken)
+    {
+        var recordingChatId = await ChatUI.RecordingChatId.Use(cancellationToken).ConfigureAwait(false);
+        if (recordingChatId.IsEmpty)
+            return (recordingChatId, 0);
+
+        var (_, end) = await Chats.GetIdRange(Session, recordingChatId, ChatEntryType.Text, cancellationToken).ConfigureAwait(false);
```
alexyakunin, line 133: 1) `ChatEntryType.Audio`?
2) If we really want to track new messages from a given user, maybe it makes sense to add `GetIdRange(session, chatId, entryType, authorId, ct)`?

And 3)... Honestly, I'd rather do this in the audio processing pipeline, not here. Here it looks more like a misuse of a feature - i.e. using what we write for the feedback loop instead of just stopping the recording in the right place.

src/dotnet/Core/StateExt.cs
```diff
@@ -2,7 +2,7 @@ namespace ActualChat;
 
 public static class StateExt
 {
-    public static Task When<T>(this IState<T> state,
+    public static Task<IComputed<T>> When<T>(this IState<T> state,
```
alexyakunin, line 5: Thanks for fixing this.

src/dotnet/Audio.Service/Processing/OpenAudioSegment.cs
```diff
@@ -47,14 +47,6 @@ public void SetRecordedAt(Moment? recordedAt)
                 Index, StreamId);
     }
 
-    public void SetAudibleDuration(TimeSpan audibleDuration)
```
alexyakunin, line 50: Why it's deleted? I.e. we are supposed to it, and if it wasn't used, we should investigate why. @alexis-kochetov is this because of VAD?

src/dotnet/Chat/Module/ChatSettings.cs
```diff
@@ -4,4 +4,6 @@ public class ChatSettings
 {
     public string Db { get; set; } = "";
     public string Redis { get; set; } = "";
+
+    public TimeSpan TurnOffRecordingAfterIdleTimeout { get; set; } = TimeSpan.FromMinutes(2);
```
alexyakunin, line 8: Should go to audio module & be `IdleRecordingTimeout`.

alexyakunin (approved):
Let's merge most of this _but_ the code that stops recording itself. IMO we need to do this differently - I left a comment on this.

src/dotnet/Chat.UI.Blazor/Services/ChatUIStateSync.cs
alexyakunin: There are a few reasons to do this in our audio pipeline:
- All such code should be in one place. Spreading across the codebase & basing it on completely different (and even unnecessary) things is deeply wrong.
- We anyway need some ticks in our audio pipeline - at least to detect audio issues, but also to run such actions as detecting idle recordings.

alexyakunin:
Sorry, just spoke w/ guys, the approach you use is the right one - yeah, it makes sense to do this based on everyone's activity in chat.

alexyakunin:
I'd still add a condition like "turn it off if current user doesn't talk for 5+ min".

src/dotnet/Chat.UI.Blazor/Services/ChatUIStateSync.cs
alexis-kochetov: Please disregard last sentences about audio pipeline - let's merge the PR

src/dotnet/Chat/Module/ChatSettings.cs
FrolYo: done

src/dotnet/Audio.Service/Processing/OpenAudioSegment.cs
FrolYo: added todo comment for @alexis-kochetov

PR #244, opened 2022-06-01
feat: db per integration test #237

### LINKED ISSUE #237: feat: stabilize/parallelize integration tests

FrolYo:
Closes #237

### DIFF
--- a/tests/Chat.IntegrationTests/ChatActivityTest.cs
+++ b/tests/Chat.IntegrationTests/ChatActivityTest.cs
@@ -13,7 +13,7 @@ public class ChatActivityTest : AppHostTestBase
     [Fact]
     public async Task BasicTest()
     {
-        using var appHost = await TestHostFactory.NewAppHost();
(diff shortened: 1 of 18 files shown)


tests/Chat.IntegrationTests/ChatActivityTest.cs
```diff
@@ -13,7 +13,7 @@ public class ChatActivityTest : AppHostTestBase
     [Fact]
     public async Task BasicTest()
     {
-        using var appHost = await TestHostFactory.NewAppHost();
+        using var appHost = await TestHostFactory.NewAppHost(Out);
```
crui3er, line 16: What if we add NewAppHost instance method to AppHostTestBase that capture Out internally, then we would be able to use shorter syntax in tests:
```
using var appHost = await NewAppHost();
```

FrolYo: done

PR #251, opened 2022-06-04
feat: keepDisplayAwake only when recording/playback is active #249

FrolYo:


src/nodejs/index.ts
```diff
@@ -10,6 +10,7 @@ export * as core from '../dotnet/UI.Blazor/exports';
 export * as audio from '../dotnet/Audio.UI.Blazor/exports';
 export * as chat from '../dotnet/Chat.UI.Blazor/exports';
 export * as users from '../dotnet/Users.UI.Blazor/exports';
+export * as global from './src/exports';
```
evgeniy-ac, line 13: why not use the `core`?
`export * as core from '../dotnet/UI.Blazor/exports';`

src/dotnet/Chat.UI.Blazor/Services/ChatUIStateSync.cs
```diff
@@ -28,6 +29,7 @@ public class ChatUIStateSync : WorkerBase
     private IChatUserSettings ChatUserSettings => _chatUserSettings ??= Services.GetRequiredService<IChatUserSettings>();
     private IChats Chats => _chats ??= Services.GetRequiredService<IChats>();
     private AudioSettings ChatSettings => _chatSettings ??= Services.GetRequiredService<AudioSettings>();
+    private IJSRuntime Js => _js ??= Services.GetRequiredService<IJSRuntime>();
```
alexyakunin, line 32: Should be `JS` - like `UI` :)

alexyakunin (approved):
Approved w/ one minor thingy ;)

src/dotnet/Chat.UI.Blazor/Services/ChatUIStateSync.cs
FrolYo: fixed

PR #275, opened 2022-06-20
fix navigate to chat on notify

alexis-kochetov:


src/dotnet/UI.Blazor/messaging-init.ts
```diff
@@ -1,5 +1,5 @@
 import { initializeApp } from 'firebase/app';
-import { getMessaging, getToken, GetTokenOptions, onMessage } from 'firebase/messaging';
+import { getMessaging, getToken, GetTokenOptions, onMessage,  } from 'firebase/messaging';
```
evgeniy-ac, line 2: `,` can be removed

PR #285, opened 2022-06-27
feat: unsubscribe outdated messaging devices #283

alexis-kochetov:


src/dotnet/Notification.Service/Notifications.Backend.cs
```diff
@@ -111,26 +117,45 @@ public virtual async Task<ImmutableArray<Symbol>> ListSubscriberIds(string chatI
...
+                        p.First.IsSuccess,
+                        p.First.Exception?.MessagingErrorCode,
+                        p.First.Exception?.HttpResponse,
+                    })
+                    .ToList();
+                var responseGroups = responses
+                    .GroupBy(x => x.MessagingErrorCode);
+                foreach (var responseGroup in responseGroups)
+                    if (responseGroup.Key == MessagingErrorCode.Unregistered) {
+                        var tokensToRemove = responseGroup
```
crui3er, line 141: It's a bit strange that NotifySubscribers method invalidates ListDevices result.
I suppose it make sense to refactor it in the following way:
1) introduce command RemoveDevicesCommand(ImmutableArray<string> deviceIds)
2) during processing response results collect unregistered device ids.
3) after all responses are processed, create RemoveDevicesCommand and invoke it.

It should be responsibility of RemoveDevicesCommand handler to invalidate ListDevices result.
Also I see possible issue with executing in background RemoveOutdatedTokens method. In fact we can have a race condition situation that ListDevices result invalidation may happen before RemoveOutdatedTokens actually executed. It should be solved eventually but still approach with RemoveDevicesCommand handler looks more predictable to me.

crui3er (approved):

PR #295, opened 2022-06-29
Navigate to message on notification

alexis-kochetov:


src/dotnet/Chat.UI.Blazor/Components/ChatView/ChatView.razor.cs
```diff
@@ -233,4 +229,17 @@ private Task OnNavigateToChatEntry(NavigateToChatEntry navigation, CancellationT
         NavigateToEntry(navigation.ChatEntryId);
         return Task.CompletedTask;
     }
+
+    private void OnLocationChanged(object? sender, LocationChangedEventArgs e)
+        => TryNavigateToEntryIfExists();
```
crui3er, line 234: OnLocationChanged is invoked once even after dispose was called when you are navigating out of chat page. I think make sense to set isDisposed flag in Dispose method and check it here. I will allow to avoid situation when we are navigating out of chat page to url with hash and it spawn avigating to chat entry on disposed component.

src/dotnet/Chat.UI.Blazor/Components/ChatView/ChatView.razor.cs
```diff
@@ -233,4 +229,17 @@ private Task OnNavigateToChatEntry(NavigateToChatEntry navigation, CancellationT
...
+
+    private void TryNavigateToEntryIfExists()
+    {
+        var currentUri = new Uri(Nav.Uri);
+        var entryIdString = currentUri.Fragment?.TrimStart('#');
+        if (long.TryParse(entryIdString, out var entryId) && entryId > 0) {
+            NavigateToEntry(entryId);
+            ChatUI.HighlightedChatEntryId.Value = entryId;
```
crui3er, line 242: I see here a room for improvement. But apparently it better to do in a separate task.
ChatUI.HighlightedChatEntryId.Value is reseted to 0L automatically in 2 seconds. If chat entry we are navigating to is not loaded yet and it takes significant amout of time to bring it to viewport and displey, it may happen that highlighing will last shorter than 2 seconds or not visible at all.

crui3er (changes requested):

src/dotnet/Notification.Service/Notifications.Backend.cs
```diff
@@ -157,6 +163,18 @@ public virtual async Task<ImmutableArray<Symbol>> ListSubscriberIds(string chatI
 
         }
 
+        async Task<string> AuthorNameResolver(string authorId)
```
FrolYo, line 166: why such naming like it's delegate or class?

alexis-kochetov: What do you suggest then

src/dotnet/Chat/Markup/Visitors/ToStringVisitor.cs
```diff
@@ -0,0 +1,110 @@
...
+    {
+        if (Builder.Length >= MaxLength)
+            return ValueTask.FromResult(Unit.Default);
+
+        Builder.Append(markup.Text);
+        return ValueTask.FromResult(Unit.Default);
+    }
+
+    protected override ValueTask<Unit> VisitUnparsed(UnparsedTextMarkup markup, CancellationToken cancellationToken)
+    {
```
FrolYo, line 103: Looks like most of methods use same code 
```csharp
        if (Builder.Length >= MaxLength)
            return ValueTask.FromResult(Unit.Default);

        Builder.Append(markup.Text);
        return ValueTask.FromResult(Unit.Default);
```

So let's extract default implementation like `VisitDefault`?

FrolYo (changes requested):

PR #317, opened 2022-07-24
feat: edit message #311

FrolYo:


src/dotnet/Notification.Service/Notifications.Backend.cs
```diff
@@ -163,15 +162,15 @@ public virtual async Task<ImmutableArray<Symbol>> ListSubscriberIds(string chatI
 
         }
 
-        async Task<string> AuthorNameResolver(string authorId)
+        async Task<string> AuthorNameResolver(string authorId, CancellationToken ct)
```
alexis-kochetov, line 165: for oarameters of nested methods we use `1` suffix - `cancellationToken1`

alexis-kochetov (approved):

src/dotnet/Notification.Service/Notifications.Backend.cs
FrolYo: done

PR #326, opened 2022-07-27
feat: virtual list with absolute positions for new items

alexis-kochetov:


.run/App.Server.run.xml
```diff
@@ -16,7 +16,7 @@
     <option name="PROJECT_WORKING_DIRECTORY_TRACKING" value="1" />
     <option name="PROJECT_KIND" value="DotNetCore" />
     <option name="PROJECT_TFM" value="net6.0" />
-    <browser url="https://localhost:7081/chat/the-actual-one" name="98ca6316-2f89-46d9-a9e5-fa9e2b0625b3" start="true" />
+    <browser url="https://local.actual.chat/chat/the-actual-one" name="98ca6316-2f89-46d9-a9e5-fa9e2b0625b3" />
```
FrolYo, line 19: why are we trying to replace localhost with local.actual.chat? does local.actual.chat support page autoreload on connection restore?

src/dotnet/Chat.UI.Blazor/Components/ChatView/ChatView.razor.cs
```diff
@@ -127,14 +123,17 @@ private async Task MonitorVisibleKeyChanges(CancellationToken cancellationToken)
...
+    async Task<VirtualListData<ChatMessageModel>> IVirtualListDataSource<ChatMessageModel>.GetData(
         VirtualListDataQuery query,
+        VirtualListData<ChatMessageModel> oldData,
         CancellationToken cancellationToken)
     {
         await WhenInitialized;
 
         var chat = Chat;
         var chatId = chat.Id.Value;
+        var currentAuthor = await ChatAuthors.Get(Session, chatId, cancellationToken);
```
FrolYo, line 135: I was told that we do not almost use `current` as a useless prefix

FrolYo (approved):

.run/App.Server.run.xml
alexis-kochetov: `local.actual.chat` uses nginx so image proxy works locally as well. Additionally notifications also works when the site is not localhost

alexis-kochetov: Why do you expect it doesn't reload on conneciton restore?

src/dotnet/Chat.UI.Blazor/Components/ChatView/ChatView.razor.cs
alexis-kochetov: OK

.run/App.Server.run.xml
FrolYo: Nginx will show you 502 error if local dev host is stopped. 502 is not automatically reloaded when host is restarted.

Anyways, I think it's such a minor inconvenience comparing to benefits you mentioned above

PR #331, opened 2022-08-01
Hand button and notify/call panel

iqmulator:


src/dotnet/Chat.UI.Blazor/Components/SlateEditor/slate-editor.ts
```diff
@@ -68,8 +68,10 @@ export class SlateEditor {
 
     public focus = () => {
         const input = this.editorDiv.querySelector('div');
+        const width = document.documentElement.clientWidth;
         if (input)
-            input.focus();
+            if (width >= 768)
```
crui3er, line 73: I think this is a bad solution. Focusing will never on wide screens with this change.
I think you can apply this condition below in onRendered method where this.autofocus flag is checking.

crui3er (changes requested):

crui3er (approved):

PR #348, opened 2022-08-06
chore: extracted KeepAwakeUI from shared module

FrolYo:


src/dotnet/UI.Blazor/keep-screen-awake.ts
```diff
@@ -1,5 +1,5 @@
 import { default as NoSleep } from 'nosleep.js';
-import { addInteractionHandler } from 'first-interaction';
+import { addInteractionHandler } from '../../nodejs/src/first-interaction';
```
alexis-kochetov, line 2: Let's remove `../../nodejs/src/`

alexyakunin: Will fix this everywhere.

PR #371, opened 2022-08-12
feat: skeletons for avatar badges #369

alexis-kochetov:


src/dotnet/UI.Blazor/Components/AuthorBadge/AuthorName.razor
```diff
@@ -0,0 +1,35 @@
+@namespace ActualChat.UI.Blazor.Components
+@using ActualChat.Users
+
+@if (ReferenceEquals(Author, Users.Author.None)) {
```
crui3er, line 4: I noticed that in AuthorCircle you also compara Author against null.
Is there a reason why you don't do it here?

alexis-kochetov: I will do the same

crui3er (approved):

Issue #389, opened 2022-08-18
Online status works unstable


FrolYo:
Users become offline at the first 2 min

PR #405, opened 2022-08-22
fix: online status #389

FrolYo:


src/dotnet/Users.Contracts/UserConstants.cs
```diff
@@ -14,4 +14,11 @@ public static class Claims
     {
         public static string Status => "urn:actual.chat:status";
     }
+
+    public static class Presence
+    {
+        // TODO: move this to configurable options when client side settings API is implemented
```
alexis-kochetov, line 20: There is local settings API - Kvas

alexis-kochetov (approved):

src/dotnet/Users.Contracts/UserConstants.cs
alexyakunin: Hi, sorry, this is done, and done a bit better in terms of constants & actual timings... I didn't look into this PR when I started, sorry. I migrated one thing that was missing in my case (extra option).

PR #407, opened 2022-08-22
context menu instead of dropdown

iqmulator:


src/dotnet/App.Wasm/wwwroot/blazorContextMenu.js
```diff
@@ -162,6 +162,16 @@ var blazorContextMenu = function (blazorContextMenu) {
 
     let showMenuCommon = function (menu, menuId, x, y, target, triggerDotnetRef) {
         let isLeftHalf = x < window.innerWidth / 2;
+        if (menuId === 'menu-AttachMenu') {
```
crui3er, line 165: Why do we need again special processing for a new menu?

crui3er (changes requested):

src/dotnet/App.Wasm/wwwroot/blazorContextMenu.js
```diff
@@ -152,28 +152,62 @@ var blazorContextMenu = function (blazorContextMenu) {
...
     };
 
-    let showMenuCommon = function (menu, menuId, x, y, target, triggerDotnetRef) {
-        let isLeftHalf = x < window.innerWidth / 2;
+    let showMenuCommon = function (menu, menuId, x, y, event, triggerDotnetRef) {
+        const target = event.target;
+        const offset = 5;
+        const isLeftHalf = x < window.innerWidth / 2;
+        const isTopHalf = y < window.innerHeight / 2;
+        if (event.button === 2) {
```
crui3er, line 168: Why do you need special processing for right click?
If you need special processing for ContextMenuToggleTrigger, then i think better to render it with additional special attribute (e.g. data-context-menu-toggle=true) and test it's value in js.

crui3er (changes requested):

crui3er (approved):

PR #422, opened 2022-08-26
right panel state

iqmulator:


src/dotnet/UI.Blazor/Components/SideNav/SideNavOptions.cs
```diff
@@ -3,4 +3,5 @@ namespace ActualChat.UI.Blazor.Components.SideNav;
 public class SideNavOptions
 {
     public SideNavDirection Direction { get; set; }
+    public bool FirstRender { get; set; }
```
crui3er, line 6: Why do you need this property? It's always true.

crui3er (approved):

PR #423, opened 2022-08-26
integration tests fix

FrolYo:


src/dotnet/Core/DI/LazyInstance.cs
```diff
@@ -0,0 +1,8 @@
+namespace ActualChat.DI;
+
+public class LazyInstance<T>: Lazy<T> where T : notnull
+{
+    public LazyInstance(IServiceProvider serviceProvider)
```
crui3er, line 5: Looks like there is no lazy behaviour. Instance of T is created upon LazyInstance<T> creation.

FrolYo: why do you think so? I pass lambda (method ref) creating instance of T to base constructor (of Lazy), not the instance itself

alexis-kochetov (approved):

src/dotnet/Core/DI/LazyInstance.cs
crui3er: Method group confused me.

PR #424, opened 2022-08-26
audio stream distribution service

alexis-kochetov:


src/dotnet/Audio.Service/AudioHub.cs
```diff
@@ -6,33 +6,48 @@ namespace ActualChat.Audio;
...
         IAudioProcessor audioProcessor,
-        AudioStreamer audioStreamer,
-        TranscriptStreamer transcriptStreamer,
+        IAudioStreamServer audioStreamServer,
+        ITranscriptStreamServer transcriptStreamServer,
         SessionMiddleware sessionMiddleware)
     {
+        AudioStreamServer = audioStreamServer;
+        TranscriptStreamServer = transcriptStreamServer;
         _audioProcessor = audioProcessor;
```
FrolYo, line 23: maybe all deps as props?

src/dotnet/Audio.Service/AudioHubBackend.cs
```diff
@@ -0,0 +1,54 @@
+using ActualChat.Transcription;
+using Microsoft.AspNetCore.SignalR;
+
+namespace ActualChat.Audio;
+
+public class AudioHubBackend : Hub
```
FrolYo, line 6: not very clear what AudioHubBackend is...

src/dotnet/Audio.Service/AudioHubBackend.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    private AudioStreamServer AudioStreamServer { get; }
+    private TranscriptStreamServer TranscriptStreamServer { get; }
+
+    public AudioHubBackend(AudioStreamServer audioStreamServer, TranscriptStreamServer transcriptStreamServer)
```
FrolYo, line 11: why class deps? in AudioHub you reference IAudioStreamServer, but here - AudioStreamServer

alexis-kochetov: Because contract interface has been implemented by Proxy - and in this case we would like to deal with actual implementation only

src/dotnet/Audio.Service/AudioHubBackend.cs
alexis-kochetov: It's a backend only Hub dedicated to cross service audio stream distribution

src/dotnet/Audio.Service/AudioHub.cs
alexis-kochetov: Sure

PR #456, opened 2022-09-01
feat: layout type provider

crui3er:
LayoutTypeProvider component publish cascasing property of type LayoutType.
LayoutType is enum that has 2 values: Desktop and Mobile. Value is set to Desktop  when window width >= 1024 px.
Example of usage see in ChatMessageEditor.razor.

### DIFF
--- a/src/dotnet/UI.Blazor/Components/LayoutTypeProvider/LayoutType.cs
+++ b/src/dotnet/UI.Blazor/Components/LayoutTypeProvider/LayoutType.cs
@@ -1,0 +1,3 @@
+namespace ActualChat.UI.Blazor.Components;
+
+public enum LayoutType { Desktop, Mobile  }
--- a/src/dotnet/UI.Blazor/Components/LayoutTypeProvider/LayoutTypeProvider.razor
+++ b/src/dotnet/UI.Blazor/Components/LayoutTypeProvider/LayoutTypeProvider.razor
@@ -1,0 +1,42 @@
+@namespace ActualChat.UI.Blazor.Components
+@using ActualChat.UI.Blazor.Module
+@implements ILayoutTypeProviderBackend
+
+@if (_layoutType.HasValue) {
+    <CascadingValue Value="@_layoutType.Value">
+        @ChildContent
+    </CascadingValue>
+}
+
+@code {
+    private LayoutType? _layoutType;
+
+    private DotNetObjectReference<ILayoutTypeProviderBackend> BlazorRef { get; set; } = null!;
+    private IJSObjectReference JSRef { get; set; } = null!;
+
+    [Inject] IJSRuntime JS { get; set; } = null!;
+    [Parameter] public RenderFragment ChildContent { get; set; } = null!;
+
+    protected override async Task OnAfterRenderAsync(bool firstRender) {
+        if (!firstRender)
+            return;
+        BlazorRef = DotNetObjectReference.Create<ILayoutTypeProviderBackend>(this);
+        JSRef = await JS.InvokeAsync<IJSObjectReference>(
+            $"{BlazorUICoreModule.ImportName}.LayoutTypeProvider.create",
+            BlazorRef
+            );
+    }
+
+    [JSInvokable("OnLayoutChanged")]
+    public void OnLayoutChanged(bool isDesktopLayout) {
+        _layoutType = isDesktopLayout ? LayoutType.Desktop : LayoutType.Mobile;
+        StateHasChanged();
+    }
+
+    public async ValueTask DisposeAsync() {
+        await JSRef.DisposeSilentlyAsync("dispose");
+        JSRef = null!;
+        BlazorRef.DisposeSilently();
+        BlazorRef = null!;
+    }
+}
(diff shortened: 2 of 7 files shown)


src/dotnet/UI.Blazor/Components/LayoutTypeProvider/LayoutType.cs
```diff
@@ -0,0 +1,3 @@
+namespace ActualChat.UI.Blazor.Components;
+
+public enum LayoutType { Desktop, Mobile  }
```
alexyakunin, line 3: What do you think about doing this a bit differently:
- `LayoutType` -> `record ScreenInfo` with property `ScreenSizeKind` (enum of `Unknown`, `Small`, `Medium`, `Large`, `ExtraLarge`, `ExtraLarge2`) -  https://tailwindcss.com/docs/responsive-design
- `IsDesktop` and `IsMobile` will be calculated based on above properties.
- Rename everything to `ScreenInfoXxx`.

alexyakunin: I kinda against `Layout` because it is a bit confusing - there is a concept of Layout in Blazor, which is different to this.

src/dotnet/UI.Blazor/Components/LayoutTypeProvider/LayoutTypeProvider.razor
```diff
@@ -0,0 +1,42 @@
+@namespace ActualChat.UI.Blazor.Components
+@using ActualChat.UI.Blazor.Module
+@implements ILayoutTypeProviderBackend
+
+@if (_layoutType.HasValue) {
+    <CascadingValue Value="@_layoutType.Value">
```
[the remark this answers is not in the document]
alexyakunin: I would say we need to assume `ScreenInfo.Size.Unknown` means e.g. `IsMobile` (assuming we render less in this case).

Alternatively, we can add `IsUnknown`, `IsMobile`, `IsDesktop`, `IsMobileOrUnknown`, `IsDesktopOrUnknown`.

src/dotnet/UI.Blazor/Components/LayoutTypeProvider/LayoutTypeProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+    private layoutLastValue : boolean;
+
+    static create(blazorRef: DotNet.DotNetObject): LayoutTypeProvider {
+        return new LayoutTypeProvider(blazorRef);
+    }
+
+    constructor(blazorRef: DotNet.DotNetObject) {
+        this.blazorRef = blazorRef;
+        this.window = document.defaultView;
+        this.window.addEventListener('resize', this.windowResizeListener);
```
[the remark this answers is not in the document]
alexyakunin: https://stackoverflow.com/questions/19669786/check-if-element-is-visible-in-dom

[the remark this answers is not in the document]
alexyakunin: https://stackoverflow.com/a/48303093 - this might be a better option.

alexyakunin (commented):
IMO makes sense to implement the suggested changes.

src/dotnet/UI.Blazor/Components/LayoutTypeProvider/LayoutType.cs
alexyakunin: Thought a bit more about this - maybe `enum ScreenSize` (no flags) with a bunch of extension methods is actually a better option, assuming:
- We set it to max available screen size (e.g. `ScreenSize.Medium`) & use just even numbers
- `Unknown` should be mapped to a value preceeding default screen size (e.g. `Medium`) & use the odd number
- `IsDesktop()`, etc. become extension methods.

alexyakunin: We can also implement `ScreenSize.IsMediumOrLarger()`, etc.

alexyakunin: + Technically we can construct it as flags too - e.g. to simplify all these checks, we just don't have to use actual `[Flags]` attribute there.

src/dotnet/UI.Blazor/Components/LayoutTypeProvider/LayoutTypeProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+        const newLayout = this.isDesktopLayout();
+        if (newLayout === this.layoutLastValue)
+            return;
+        this.layoutLastValue = newLayout;
+        this.notifyLayoutChanged(newLayout);
+    });
+
+    private notifyLayoutChanged(isDesktop : boolean)
+    {
+        this.blazorRef.invokeMethodAsync('OnLayoutChanged', isDesktop);
```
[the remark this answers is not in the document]
crui3er: It's already does. See callee.

PR #459, opened 2022-09-02
fix mobile panel behavior

iqmulator:


src/dotnet/Chat.UI.Blazor/Components/ChatMessageEditor/chat-message-editor.ts
```diff
@@ -29,19 +31,73 @@ export class ChatMessageEditor {
...
             attributes: true,
         });
         this.changeMode();
     }
 
+    private getSlateInput = () : HTMLDivElement => {
+        this.slateInput = this.input.querySelector('[role="textbox"]');
+        if (this.slateInput)
+            return this.slateInput;
+        throw new Error("Slate editor not found.");
```
alexyakunin, line 60: Maybe just log the error here, if its absence doesn't break things?

src/dotnet/Chat.UI.Blazor/Components/ChatMessageEditor/chat-message-editor.ts
```diff
@@ -29,19 +31,73 @@ export class ChatMessageEditor {
         this.notifyPanel = this.editorDiv.querySelector(':scope div.post-panel .notify-call-panel');
 
         // Wiring up event listeners
+        this.editorDiv.addEventListener('click', this.editorFocusHandler);
```
alexyakunin, line 34: Ok... I kinda tired talking about this:
- all event handlers should have "onXxx" name pattern
- moreover, you invented another kind here: all other handlers are called "xxxListener", and you add "xxxHandler".

How can we stop this?

alexyakunin (changes requested):
LGTM with minor stuff.

PR #466, opened 2022-09-05
fix: contact list - highlight search phrase #462

FrolYo:


src/dotnet/Chat.UI.Blazor/Components/ContactNavItem.razor
```diff
@@ -68,4 +68,5 @@
             isRecording);
     }
 
+    public record Model(bool IsOnline, string ActiveChatId, string FullChatId, string ShortChatId, bool IsListening, bool IsRecording);
```
alexyakunin, line 71: Let's split this to lines.

alexyakunin (changes requested):

src/dotnet/Chat.UI.Blazor/Components/ContactNavItem.razor
alexyakunin: One more thing: let's also order this stuff + do we really need "ActiveChatId" here? I.e. why not `IsActive`?

alexyakunin: All in all, let's make this readable, logically structured, + w/o unnecessary data.

PR #471, opened 2022-09-05
feat: search in chat #469

FrolYo:


src/dotnet/Chat.UI.Blazor/Components/ChatSearch.razor
```diff
@@ -0,0 +1,28 @@
...
+
+    [CascadingParameter] public Chat Chat { get; set; } = null!;
+
+    private async Task OnSearchTriggered(string text) {
+        var chatId = Chat.Id;
+        var nextEntry = await Chats.FindNext(Session, chatId, _lastFoundEntry?.Id, text, CancellationToken.None);
+        if (nextEntry == null)
+            return;
+
+        await UIEventHub.Publish(new NavigateToChatEntry(nextEntry.Id));
```
alexyakunin, line 20: Btw, IMO this is a misuse of events. I know it's not your design, but IMO this should be simply a method somewhere in `ChatUI`, which also checks if the current chat matches, etc.

src/dotnet/Chat.UI.Blazor/Components/ChatSearch.razor
```diff
@@ -0,0 +1,28 @@
...
+        ChatUI.HighlightedChatEntryId.Value = nextEntry.Id;
+        _lastFoundEntry = nextEntry;
+    }
+
+    private void OnSearchCancelled() {
```
alexyakunin, line 25: Can we start using expression bodied members when appropriate?

src/dotnet/Chat.UI.Blazor/Components/ContactListNavbarWidget.razor
```diff
@@ -130,7 +130,7 @@
             case ContactGroup.Favorites:
```
alexyakunin, line 130: Not related to this PR, but IMO we should either have pins or favorites. Let's talk about this once AK is back & decide what's better here.

src/dotnet/Core/Constants.cs
```diff
@@ -10,7 +10,7 @@ public static class Chat
         public static TileStack<Moment> TimeTileStack { get; } = TileStacks.Moment3MTo6Y;
         public static TimeSpan MaxEntryDuration { get; } = TimeTileStack.MinTileSize.EpochOffset; // 3 minutes, though it can be any
         public const int PictureFileSizeLimit = 10 * 1024 * 1024; // 10MB
-        public const int MaxRecentPeerChats = 5;
+        public const int MaxSearchCriteriaLength = 100;
```
alexyakunin, line 13: If I am not mistaken, I started to call it "search filter".

It's mostly because of two things:
- Shorter is better
- "Filter" is kinda less broad than "Criteria" (criteria can be arbitrary complex - i.e. all kinds of fields you search by, etc., filter is perceived as something that's simpler).

src/dotnet/Core/Text/StringExt.cs
```diff
@@ -19,6 +19,8 @@ public static string RequireNonEmpty(this string? source, string name)
     [return: NotNullIfNotNull("source")]
     public static string? RequireEmpty(this string? source, string name)
         => source.IsNullOrEmpty() ? source : throw StandardError.Constraint($"{name} must be null or empty here.");
+    public static string? RequireMaxLength(this string source, int length, string name)
+        => source.Length <= length ? source : throw StandardError.Constraint($"{name} must not be longer than {length}.");
```
alexyakunin, line 23: Maybe a better option: "Must be no more than X characters." (checked Google, 38M vs 1.4M results)

alexyakunin (approved):
Check out my comments & let's merge this to Dev.

You can optionally cover triggering of this feature with IncompleteUI flag for now - i.e. let's polish it on Dev first, esp. UI & interaction.

alexyakunin:
Btw, overall, cool stuff - i.e. yeah, this is definitely super handy.

src/dotnet/Core/Constants.cs
FrolYo: done

src/dotnet/Core/Text/StringExt.cs
FrolYo: done

src/dotnet/Chat.UI.Blazor/Components/ChatSearch.razor
FrolYo: done

PR #474, opened 2022-09-06
new mobile panel behavior

iqmulator:


alexis-kochetov (commented):
I think it make sense to use https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API

src/dotnet/Chat.UI.Blazor/Components/ChatMessageEditor/chat-message-editor.ts
```diff
@@ -1,5 +1,6 @@
 import './chat-message-editor.css';
 import { SlateEditorHandle } from '../SlateEditor/slate-editor-handle';
+import { debounce } from 'lodash';
```
alexis-kochetov, line 3: We already have `debounce`

PR #487, opened 2022-09-07
fix: push notification messages

FrolYo:


src/dotnet/Chat/Markup/Mention.cs
```diff
@@ -14,6 +14,7 @@ public override string Format()
 
     public string Format(MentionFormat format)
         => format switch {
+            MentionFormat.NameOnly => $"@{Name.NullIfEmpty() ?? Id}",
```
alexyakunin, line 17: Check out `_` clause - it is actually responsible for this.

src/dotnet/Chat/Markup/Visitors/MarkupTrimmer.cs
```diff
@@ -74,7 +74,7 @@ protected override Markup VisitCodeBlock(CodeBlockMarkup markup)
         if (sb.Length == 0)
             return AppendEnd();
         Append(sb.Length);
-        return new MarkupSeq(markup with { Code = sb.ToString() }, AppendEnd());
+        return new MarkupSeq(markup with { Code = sb.ToString() });
```
alexyakunin, line 77: You shouldn't remove `AppendEnd` call here - see the first `if`, we need to append end in this case, coz we're trimming the code fragment.

src/dotnet/Chat/Markup/Visitors/MarkupTrimmer.cs
```diff
@@ -83,9 +83,7 @@ protected override Markup VisitText(TextMarkup markup)
             return AppendEnd();
         markup = markup with { Text = markup.Text.Truncate(MaxLength - Length) };
         Append(markup.Text.Length);
-        return new MarkupSeq(
-            markup with { Text = markup.Text.Truncate(MaxLength - Length) },
-            AppendEnd());
+        return new MarkupSeq(markup with { Text = markup.Text.Truncate(MaxLength - Length) });
```
alexyakunin, line 86: Same here.

src/dotnet/Chat/Markup/Mention.cs
alexyakunin: For the note, NameOnly means we never show Ids, so my code in the end is right. IMO it's fine to quote the string all the times here for now, btw, since the names are going to be with spaces, etc. - in short, let's solve nice quoting later.

alexyakunin (changes requested):
Please fix the issues.

alexyakunin:
If you want, you can add tests for some of these trimming cases (e.g. for code).

src/dotnet/Chat/Markup/Mention.cs
FrolYo: fixed

src/dotnet/Chat/Markup/Visitors/MarkupTrimmer.cs
FrolYo: fixed

src/dotnet/Chat/Markup/Visitors/MarkupTrimmer.cs
FrolYo: fixed

alexis-kochetov (approved):

FrolYo:
Done

PR #505, opened 2022-09-12
fix: initial recent entry date #495

FrolYo:


alexyakunin:
LGTM, merging now, since this seems important.

PR #507, opened 2022-09-13
Notifications refactoring + Command.Enqueue + IEvent

alexis-kochetov:


src/dotnet/Chat.Service/ChatsBackend.cs
```diff
@@ -415,10 +414,13 @@ await ownerUserIds
...
-        if (!entry.Content.IsNullOrEmpty() && !entry.IsStreaming && entry.Type == ChatEntryType.Text) {
-            var chatEvent = new NewChatEntryEvent(entry.ChatId, entry.Id, entry.AuthorId, entry.Content);
-            await EventPublisher.Publish(chatEvent, cancellationToken).ConfigureAwait(false);
-        }
+        if (!entry.Content.IsNullOrEmpty() && !entry.IsStreaming && entry.Type == ChatEntryType.Text)
+            await new NewTextEntryEvent(entry.ChatId, entry.Id, entry.AuthorId, entry.Content)
+                .Configure()
+                .ShardByChatId(entry.ChatId)
+                .WithPriority(EventPriority.Low)
+                .ScheduleOnCompletion(command)
```
alexyakunin, line 422: I'd use this syntax: `.EnqueueOnCompletion(QueueRef queueRef)`. 

`QueueRef` would typically be constructed as: `Queues.Chats.Shard(chatId).Priority(...)`.

alexyakunin: As far as I can judge, async API isn't absolutely necessary here assuming we just put this info to e.g. `commandContext.Outermost.Items` & process them either right before or right after the commit.

And if it's going to be async (this is more flexible), I'd expose it as ValueTask for now, + add CancellationToken arg.

src/dotnet/Chat.Service/Jobs/ChatJobs.cs
```diff
@@ -1,54 +1,56 @@
...
         await JoinToAnnouncementsChat(@event.UserId, cancellationToken).ConfigureAwait(false);
 
-        if (_hostInfo.IsDevelopmentInstance)
+        if (HostInfo.IsDevelopmentInstance)
             await JoinAdminToDefaultChat(@event.UserId, cancellationToken).ConfigureAwait(false);
     }
 
+
+
     private async Task JoinToAnnouncementsChat(string userId, CancellationToken cancellationToken)
```
alexyakunin, line 40: `JoinAnnouncementsChat`

src/dotnet/Chat.Service/Jobs/ChatJobs.cs
```diff
@@ -57,12 +59,12 @@ private async Task JoinToAnnouncementsChat(string userId, CancellationToken canc
 
     private async Task JoinAdminToDefaultChat(string userId, CancellationToken cancellationToken)
```
alexyakunin, line 60: JoinDefaultChatIfAdmin

src/dotnet/Chat.Service/Module/ChatServiceModule.cs
```diff
@@ -100,10 +98,14 @@ public override void InjectServices(IServiceCollection services)
         services.AddResponseCaching();
```
alexyakunin, line 98: Btw, do we use this option?

alexyakunin: I guess just for the static assets.

alexyakunin (commented):
Did a very partial review for now (ended where the last comment is), will do more detailed one later.

src/dotnet/Chat.Service/ChatsBackend.cs
alexyakunin: Why `Queues`? Well, we can statically store all the builders there, which are structs. The builders themselves aren't doing anything but provide extension methods, + once completed, they produce `QueueRef`, which has just `string QueueId` property.

Maybe we don't even need `QueueRef` - i.e. we can agree on `.EnqueueOnCompletion(string queueId, CancellationToken cancellationToken)` syntax too. The key thing here is how they're built.

One other interesting q. is: should this QueueRef be a part of event? IMO not, coz as you've said, technically it might be routed to N queues, which also means `.EnqueueOnCompletion` may have an overload with `string[] queueIds`.

alexyakunin: Also, on names:
- `ScheduleXxx` usually means "run at certain time / with certain delay / periodically / etc."
- `EnqueueXxx` means "add to queue / run on one of queue processors as soon as possible".

IMO for scheduling (running with a certain delay) we'll need something else later, so I'd avoid this name for now.

PR #576, opened 2022-10-05
Feat/android notifications

crui3er:
This PR brings notifications to Android app.
It works but so far there are 2 main issues that I plan to resolve later:
1) when you tap on a notification when app is background, it can be that app is not navigated to the message correspondent to a notification.
2) it might happen that notifications are not displayed when app is kille/swipe out
there are a lot compains with such issue
https://github.com/firebase/quickstart-android/issues/368
https://github.com/firebase/quickstart-android/issues/816
May be behavior in production is different according to https://stackoverflow.com/questions/48642423/firebase-push-notification-issue-in-android-when-app-is-closed-killed
I will continue my investigation

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -161,6 +161,8 @@
 
     <!-- Android -->
     <PackageReference Update="Xamarin.GooglePlayServices.Auth" Version="120.2.0.1" />
+    <PackageReference Update="Xamarin.Firebase.Messaging" Version="123.0.8" />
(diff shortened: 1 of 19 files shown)


alexis-kochetov:
"uses-permission android:name="com.google.android.c2dm.permission.RECEIVE"
"uses-permission android:name="android.permission.WAKE_LOCK"

Have you tried?

Issue #406, opened 2022-08-22
Online status must consider user activity


FrolYo:
Current implementation of online presence status simply keeps online while app page is open.
This logic must be more accurate and take into account such user actions as clicks, typing, etc

PR #637, opened 2022-10-28
chore: chat authors refactoring #406

FrolYo:


alexyakunin:
Oh... This definitely overlaps a lot with what I do...

PR #653, opened 2022-11-06
feat: reactions notifications #644

FrolYo:


src/dotnet/Chat.Contracts/Events/ReactionChangedEvent.cs
```diff
@@ -0,0 +1,13 @@
+using ActualChat.Commands;
+
+namespace ActualChat.Chat.Events;
+
+[DataContract]
+public record ReactionChangedEvent(
+    [property: DataMember(Order = 0)] string ChatEntryId,
+    [property: DataMember(Order = 2)] string AuthorId,
```
alexis-kochetov, line 8: Looks like an error!

alexis-kochetov (changes requested):

src/dotnet/Chat.Contracts/Events/ReactionChangedEvent.cs
FrolYo: fixed

PR #663, opened 2022-11-07
fix: reaction emoji is small for large count #66

FrolYo:


alexyakunin:
IMO nothing to review, ok to merge )

PR #732, opened 2022-11-30
New landing

iqmulator:


alexis-kochetov (changes requested):
Please use SVG instead of PNG for every image except screenshots!

PR #752, opened 2022-12-14
fix: system entry in news #751

FrolYo:


src/dotnet/Chat.UI.Blazor/Services/ChatUI.cs
```diff
@@ -165,6 +167,24 @@ public virtual async Task<ImmutableList<ChatInfo>> List(ChatListOrder order, Can
...
+            return entry.GetContentOrDescription();
+
+        var membersChanged = entry.SystemEntry?.MembersChanged;
+        if (membersChanged != null) {
+            var author = await Authors.Get(Session, entry.ChatId, membersChanged.AuthorId, cancellationToken).ConfigureAwait(false);
+            var action = membersChanged.HasLeft ? "left" : "joined";
+            return $"{author.Avatar.Name} {action} the group";
+        }
+
+        throw StandardError.Constraint("Unexpected system entry type");
```
crui3er, line 185: I think it's too strict behavior. Do not throw exception here.
Better to have empty line rather than failed app.

crui3er (changes requested):

src/dotnet/Chat.UI.Blazor/Services/ChatUI.cs
FrolYo: done

alexyakunin:
This is not going to work... I'll address this.

alexyakunin:
Fixed, this PR can be closed w/o merging, @FrolYo

PR #796, opened 2023-01-09
fix: show original in new tab for local storage #795

FrolYo:


src/dotnet/Core/Blobs/Internal/LocalFolderBlobStorage.cs
```diff
@@ -1,27 +1,33 @@
...
 using Stl.IO;
 
 namespace ActualChat.Blobs.Internal;
 
 internal class LocalFolderBlobStorage : IBlobStorage
 {
-    private readonly FilePath _directory;
+    private IContentTypeProvider? _contentTypeProvider;
+    private FilePath Directory { get; }
+    private IServiceProvider ServiceProvider { get; }
```
alexyakunin, line 11: Let's be consistent & call it Services.

src/dotnet/Core/Blobs/Internal/LocalFolderBlobStorage.cs
```diff
@@ -1,27 +1,33 @@
...
+using Microsoft.AspNetCore.StaticFiles;
 using Stl.IO;
 
 namespace ActualChat.Blobs.Internal;
 
 internal class LocalFolderBlobStorage : IBlobStorage
 {
-    private readonly FilePath _directory;
+    private IContentTypeProvider? _contentTypeProvider;
+    private FilePath Directory { get; }
```
alexyakunin, line 10: The agreement is:
- Resolved services are always props
- Everything else is usual (i.e. can be a private field, property, etc. -whatever is reasonable).

And here it's reversed: _contentTypeProvider must be a property, Directory must be a field.

src/dotnet/Core/Blobs/Internal/LocalFolderBlobStorage.cs
```diff
@@ -50,8 +56,8 @@ public async Task Write(string path, Stream dataStream, string contentType, Canc
 
         ValidatePath(path);
 
-        var fullPath = _directory & path;
-        Directory.CreateDirectory(fullPath.DirectoryPath);
+        var fullPath = Directory & path;
+        System.IO.Directory.CreateDirectory(fullPath.DirectoryPath);
```
alexyakunin, line 60: Maybe makes sense to use another name in such cases - e.g. RootDirectory, etc.

alexyakunin (changes requested):

src/dotnet/Core/Blobs/Internal/LocalFolderBlobStorage.cs
FrolYo: done

src/dotnet/Core/Blobs/Internal/LocalFolderBlobStorage.cs
FrolYo: [discussed] now it's good

src/dotnet/Core/Blobs/Internal/LocalFolderBlobStorage.cs
FrolYo: done

PR #801, opened 2023-01-10
Refactor audio processor to speedup transcription response

alexis-kochetov:


src/dotnet/Audio.Service/AudioProcessor.cs
```diff
@@ -84,66 +84,66 @@ public AudioProcessor(Options settings, IServiceProvider services)
                 TimeSpan.Zero,
                 AudioSourceLog,
                 cancellationToken);
-            var openSegment = new OpenAudioSegment(0, record, audio, author, languages, OpenAudioSegmentLog);
+            var openSegment = new OpenAudioSegment(0,
```
alexyakunin, line 87: I ended up putting this on one line at some point, coz cut something & it was fitting. Sorry :)

src/dotnet/Audio.Service/AudioProcessor.cs
```diff
@@ -84,66 +84,66 @@ public AudioProcessor(Options settings, IServiceProvider services)
...
             var audioEntryTask = BackgroundTask.Run(
-                () => CreateAudioEntry(openSegment, cancellationToken),
+                () => CreateAudioEntry(openSegment, recordedAt, cancellationToken),
                 Log,
                 $"{nameof(CreateAudioEntry)} failed",
                 cancellationToken);
 
             var transcribeTask = BackgroundTask.Run(
-                () => TranscribeAudio(record.ChatId, openSegment, audioEntryTask, cancellationToken),
+                () => TranscribeAudio(record.ChatId, openSegment, audioEntryTask, CancellationToken.None),
```
alexyakunin, line 113: IMO makes sense to still pass CT here, but process all of them gracefully by shutting down source audio stream once the originally passed token gets cancelled.

alexyakunin: Maybe we should actually using "wrapped" `audioStream` - i.e. call something after `GetFrames` that will wrap it into a normally terminating stream on cancellation.

src/dotnet/Audio.Service/AudioProcessor.cs
```diff
@@ -84,66 +84,66 @@ public AudioProcessor(Options settings, IServiceProvider services)
...
-            }
+            // we don't care much about transcription errors - basically we should finalize audio entry before
+            await transcribeTask.ConfigureAwait(false);
         }
         catch (Exception e) when (e is not OperationCanceledException) {
             Log.LogError(e, "Error processing audio stream {StreamId}", streamId);
             throw;
         }
+        catch (Exception e) {
+            Log.LogWarning(e, "Cancelled processing audio stream {StreamId}", streamId);
```
alexyakunin, line 144: I usually use "failed" phrasing in logging - i.e. `"Audio processing failed for stream #{StreamId}"`.

src/dotnet/Audio.Service/AudioProcessor.cs
```diff
@@ -211,24 +211,10 @@ private async Task<Language> GetTranscriptionLanguage(AudioRecord record, Cancel
 
     private async Task<ChatEntry> CreateAudioEntry(
         OpenAudioSegment audioSegment,
+        Moment recordedAt,
         CancellationToken cancellationToken)
     {
         var now = Clocks.SystemClock.Now;
-        var beginsAt = now;
```
alexyakunin, line 217: Why this thing is removed? Seems useful.

alexyakunin: Or maybe it's not. Let's discuss this later today.

src/dotnet/Audio.Service/Processing/TranscriptSplitter.cs
```diff
@@ -5,10 +5,10 @@ namespace ActualChat.Audio.Processing;
 
 public class TranscriptSplitter : TranscriptionProcessorBase
 {
-    public static TimeSpan TextEntryWaitDelay { get; set; } = TimeSpan.FromSeconds(0.2);
-    public static float QuickSplitPauseDuration { get; set; } = 0.1f;
-    public static float SplitPauseDuration { get; set; } = 0.75f;
-    public static float SplitOverlap { get; set; } = 0.25f;
+    public static TimeSpan TextEntryWaitDelay { get; } = TimeSpan.FromSeconds(0.2);
```
alexyakunin, line 8: This should actually go to `Options` at some point :)

src/dotnet/Audio.Service/StreamServerBase.cs
```diff
@@ -7,8 +7,8 @@ public abstract class StreamServerBase<TItem> : IDisposable
 
     protected int StreamBufferSize { get; init; } = 64;
     protected TimeSpan MaxStreamDuration { get; init; } = TimeSpan.FromSeconds(600);
-    protected TimeSpan ReadStreamWaitDuration { get; init; } = TimeSpan.FromSeconds(2);
-    protected TimeSpan ReadStreamExpiration { get; init; } = TimeSpan.FromSeconds(5);
+    protected TimeSpan ReadStreamWaitDuration { get; init; } = TimeSpan.FromSeconds(10);
```
alexyakunin, line 10: Why these are tweaked? IMO such things should be tweaked only if we see real problems with the old settings.

2s wait time is there, coz it returns "no stream" only after spending this wait time. So if client sees the transcript w/ StreamId & hits the server, it shouldn't wait a lot to discover there is no stream, I guess.

alexyakunin: Let's discuss this as well. Maybe 5s is fine assuming ReadStreamExpiration is 15s.

src/dotnet/Audio.Service/StreamServerBase.cs
```diff
@@ -40,17 +40,23 @@ protected async Task<IAsyncEnumerable<TItem>> Read(Symbol streamId, Cancellation
...
-            .ContinueWith(
-                _ => entry.Dispose(),
-                CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
+
+        BackgroundTask.Run(async () => {
+            await memoizer.WriteTask.ConfigureAwait(false);
+            await Clocks.CpuClock.Delay(ReadStreamWaitDuration, CancellationToken.None).ConfigureAwait(false);
+            entry.Dispose();
+        },
+            Log,
```
alexyakunin, line 55: Something is off with formatting here.

alexyakunin: Maybe just put `}, Log, ...` on the same line?

src/dotnet/Audio.Service/StreamServerBase.cs
```diff
@@ -40,17 +40,23 @@ protected async Task<IAsyncEnumerable<TItem>> Read(Symbol streamId, Cancellation
...
+            $"{nameof(Write)} failed",
```
alexyakunin, line 56: Yep, that's how "failed" phrasing looks like :)

src/dotnet/Audio/AudioSourceExt.cs
```diff
@@ -10,6 +10,14 @@ public static AudioSource Concat(this AudioSource left, AudioSource right, Cance
             left.Log,
             cancellationToken);
 
+    public static AudioSource ConcatUntil(this AudioSource left, AudioSource right, TimeSpan duration, CancellationToken cancellationToken)
```
alexyakunin, line 13: Nice wrapper. We can add a similar one for graceful cancellation.

src/dotnet/Transcription.Service/Google/GoogleTranscriber.cs
```diff
@@ -4,129 +4,206 @@
...
+            },
+        };
+        Log.LogDebug("Starting recognize process for {StreamId}", streamId);
+        await recognizeRequests.WriteAsync(new StreamingRecognizeRequest {
+            StreamingConfig = streamingRecognitionConfig,
+            Recognizer = recognizerName,
+        }).ConfigureAwait(false);
+
+        var webMStreamAdapter = new WebMStreamAdapter(Log);
+        var silenceAudio = await _silenceAudioSourceTask.ConfigureAwait(false);
```
alexyakunin, line 71: `silenceAudioSource`?

src/dotnet/Transcription.Service/Google/GoogleTranscriber.cs
```diff
@@ -4,129 +4,206 @@
...
+        var audioSourceJoined = silenceAudio
```
alexyakunin, line 72: `resultAudioSource`?

src/dotnet/Transcription.Service/Google/GoogleTranscriber.cs
```diff
@@ -4,129 +4,206 @@
...
+            .Take(TimeSpan.FromMilliseconds(2000), cancellationToken)
+            .Concat(audioSource, cancellationToken)
+            .ConcatUntil(silenceAudio, TimeSpan.FromSeconds(7), cancellationToken);
```
alexyakunin, line 75: This overload is actually weird - `TimeSpan.FromSeconds(7)` here reads as total length, I'd probably leave just `Concat` overload with this parameter, assuming we name params there as `this head, tail, maxTailDuration`.

alexyakunin (commented):
Reviewed 50%, will do the rest a bit later.

src/dotnet/Audio.Service/AudioProcessor.cs
alexis-kochetov: We usually do not write cancellation errors  - and here I did that for debugging purposes - so it's not a failure

src/dotnet/Audio.Service/AudioProcessor.cs
alexis-kochetov: I have inlined several variables

src/dotnet/Audio.Service/StreamServerBase.cs
alexis-kochetov: in tests we have   ```
523 info: ActualChat.Audio.AudioStreamProxy[0]
      Write(audio #01GPJD65Q8QDN211QFCCQ1R7DH-0000): fallback to the local server
```
and ```
 3058 info: ActualChat.Audio.AudioStreamProxy[0]
      Read(audio #01GPJD65Q8QDN211QFCCQ1R7DH-0000): fallback to the local server
```
so 2 sec is not enough

alexis-kochetov:
```
  531 info: Stl.Fusion.Operations.Internal.CompletionProducer[0]
      Local operation completion succeeded. Agent: 'LEGION-jQBv1ZSkOkOr05O8Ijswkg-2', Command: UpsertEntryCommand 
```
Entry has been created in 10 ms, but our entry reader hasn't reacted during next 2500 ms

alexis-kochetov: At least in tests

src/dotnet/Transcription.Service/Google/GoogleTranscriber.cs
alexis-kochetov: Because this param doesn't limit the tail - but the whole stream - so it works similar to PadRight

src/dotnet/Audio.Service/StreamServerBase.cs
alexis-kochetov: I have reverted back those timings  - we just had a bug with test itself

PR #864, opened 2023-01-25
chore: ChatUI segregation #815

FrolYo:


src/dotnet/Chat.UI.Blazor/Services/AudioUI.cs
```diff
@@ -1,26 +1,102 @@
 using ActualChat.Audio;
 using ActualChat.Audio.UI.Blazor.Components;
+using ActualChat.UI.Blazor.Services;
 
 namespace ActualChat.Chat.UI.Blazor.Services;
 
-// TODO: extract into a separate class RecordingUI
-public partial class ChatUI
+public class AudioUI : WorkerBase
```
alexis-kochetov, line 7: Is it possible to move this AudioUI to Audio.UI.Blazor.Services without adding reference to Chats\Chat.UI?

alexis-kochetov (approved):

src/dotnet/Chat.UI.Blazor/Services/AudioUI.cs
FrolYo: No, cause it depends on ActiveChatsUI

PR #873, opened 2023-01-26
feat: new chat modal

evgeniy-ac:
image


src/dotnet/Chat.UI.Blazor/Components/LeftPanel/ThinLeftPanel.razor
```diff
@@ -8,7 +9,7 @@
         </div>
         <div class="flex-y w-full grow items-center gap-y-3 md:gap-y-2">
             <ButtonSquare
-                Click="OnCreateButtonClick"
+                Click="OnCreateNewChat"
```
FrolYo, line 12: as discussed offline, we stick to On-Component-Action convention
so it would be something like `OnCreateChatClick` or `OnCreateChatButtonClick`

src/dotnet/Chat.UI.Blazor/Components/LeftPanel/ThinLeftPanel.razor
```diff
@@ -25,6 +26,6 @@
 @code {
     [Inject] private ModalUI ModalUI { get; init; } = null!;
 
-    private void OnCreateButtonClick()
-        => ModalUI.Show(new CreationModal.Model());
+    private void OnCreateNewChat()
```
FrolYo, line 29: Why don't we use Task as a return type of event handler since ModalUI.Show returns Task and blazor supports Task?

FrolYo (changes requested):

src/dotnet/Chat.UI.Blazor/Components/LeftPanel/ThinLeftPanel.razor
evgeniy-ac: Done

src/dotnet/Chat.UI.Blazor/Components/LeftPanel/ThinLeftPanel.razor
evgeniy-ac: Done

PR #1036, opened 2023-03-03
fix: navigate to end does not disappear #1035

FrolYo:


src/dotnet/UI.Blazor/Components/VirtualList/virtual-list.ts
```diff
@@ -342,12 +342,11 @@ export class VirtualList {
             if (!key) {
                 if (this._endAnchorRef === itemRef) {
                     if (entry.isIntersecting) {
+                        hasChanged ||= !this._isEndAnchorVisible;
                         this._isEndAnchorVisible = true;
                         if (rs.hasVeryLastItem) {
                             const edgeKey = this.getLastItemKey();
-                            if (this._isEndAnchorVisible) {
```
crui3er, line 348: Why do you think that this condition is too strict? As I understand it means we see last item and end anchor and therefore we can set sticky edge.

crui3er: And I suppose NavigateToEnd button depends only end anchor visibility. How is it connected with sticky edge?

PR #1093, opened 2023-03-25
feat: disable non admins to access test pages #1092

FrolYo:


alexyakunin:
Could a common layout be a better solution?

PR #1119, opened 2023-04-04
Separators and updated loading screen

iqmulator:


src/dotnet/Chat.UI.Blazor/Components/ChatView/chat-view.css
```diff
@@ -109,10 +109,11 @@ body.hoverable .chat-header-title:hover {
 }
 
 .new-separator {
```
crui3er, line 111: Where is 'new-separator' css class used? Separator inside ChatMessage no longer uses it.

iqmulator: Yes, i forgot to delete. Thanks.

PR #1143, opened 2023-04-13
guides for all platforms

iqmulator:


alexis-kochetov (changes requested):
Please don't use *.svg format for raster images, use WebP instead!

alexis-kochetov (approved):

crui3er (changes requested):

PR #1196, opened 2023-05-12
close #1195 It should be possible to define whether guest user can join chat

### LINKED ISSUE #1195: It should be possible to define independently whether guest user can join chat and whether signed-in user can join chat anonymously

crui3er:
It should be possible to define independently whether guest user can join chat and whether signed-in user can join chat anonymously #1195

image

Toggles visibility are controlled by ExperimentalFeature feature toggle. So they are shown in development environment and to Grigory and admins in other environments.

Migration just adds new props to DbChat. Default values are False.
Possibility to join with guest user or anonymously should be enabled explicitly to a new and existent chats.


alexyakunin:
Public chat - anyone can join by navigating to its URL
- Visitors can join this chat without signing in [let's make it pink + add a tooltip "Spam messages are more likely if this option is enabled"
Allow users to join this chat anonymously [toolpip: Signed-in users can join this chat without revealing their identity]
- Require anonymous join [tooltip: It won't be possible to join this chat normally, so everyone's identity will be hidden here]

PR #1246, opened 2023-05-29
embedded chat and test page

alexis-kochetov:


src/dotnet/Chat.Service/ChatsBackend.cs
```diff
@@ -63,6 +63,24 @@ public ChatsBackend(IServiceProvider services) : base(services)
...
+    public virtual async Task<Chat?> GetTemplatedChatFor(ChatId templateId, UserId userId, CancellationToken cancellationToken)
+    {
+        if (templateId.IsNone)
+            throw new ArgumentOutOfRangeException(nameof(templateId));
+        if (userId.IsNone)
+            throw new ArgumentOutOfRangeException(nameof(userId));
+
+        var dbContext = CreateDbContext();
+        await using var _ = dbContext.ConfigureAwait(false);
+        var dbChat = await dbContext.Chats.AsQueryable()
```
crui3er, line 76: Why do you `.AsQueryable()` here?

src/dotnet/Chat.Service/Chats.cs
```diff
@@ -304,6 +308,140 @@ async Task Remove(ChatEntryId entryId1)
...
+        }
+        var createAuthorCommand = new IAuthorsBackend.UpsertCommand(
+            chatId,
+            AuthorId.None,
+            account.Id,
+            ExpectedVersion: null,
+            new AuthorDiff {
+                AvatarId = guestAvatar.Id,
+            });
+        await Commander.Run(createAuthorCommand, cancellationToken).ConfigureAwait(false);
```
crui3er, line 440: Is it possible to replace 'join guest author' block with IAuthors.JoinCommand?

src/dotnet/Chat.Service/ChatsBackend.cs
```diff
@@ -675,6 +712,22 @@ private async Task JoinDefaultChatIfAdmin(UserId userId, CancellationToken cance
...
     }
 
+    private async Task JoinFeedbackTemplateChatIfAdmin(UserId userId, CancellationToken cancellationToken)
+    {
+        var account = await AccountsBackend.Get(userId, cancellationToken).ConfigureAwait(false);
+        if (account is not { IsAdmin: true })
+            return;
+        if (account.Email.IsNullOrEmpty())
+            return;
+        if (!account.Email.OrdinalEndsWith("@actual.chat"))
```
crui3er, line 722: I found that we also get email with the following expression `var email = account.User.GetEmail();`
I think this variant is better to use because AccountFull.Email is not verified at the moment.
So everyone can claim that they are from actual.chat domain.
What do you think about?

src/dotnet/Chat.Service/ChatsUpgradeBackend.InitializeData.cs
```diff
@@ -358,4 +358,124 @@ string ConvertOldTextToTimeMap(string textToTimeMapJson)
...
+        var userIdByEmail = new Dictionary<string, UserId>(StringComparer.OrdinalIgnoreCase);
+        foreach (var userId in userIds) {
+            var account = await AccountsBackend.Get(userId, cancellationToken).ConfigureAwait(false);
+            if (account == null)
+                continue;
+
+            var user = account.User;
+            if (user.Claims.Count == 0)
+                continue;
+            if (!user.Claims.TryGetValue(ClaimTypes.Email, out var email))
```
crui3er, line 387: There is an extension method `GetEmail()`, better to use it

src/dotnet/Chat.Service/ChatsUpgradeBackend.InitializeData.cs
```diff
@@ -358,4 +358,124 @@ string ConvertOldTextToTimeMap(string textToTimeMapJson)
...
+                continue;
+
+            if (hostInfo.IsDevelopmentInstance) {
+                if (email.OrdinalIgnoreCaseEndsWith("actual.chat"))
+                    userIdByEmail.Add(email, userId);
+            }
+            else {
+                if (OrdinalIgnoreCaseEquals(email, "alex.yakunin@actual.chat")
```
crui3er, line 395: May be better to extract your emails in constants. There is already another place with them. See ChatsUpgradeBackend.CreateAnnouncementsChat

src/dotnet/Chat.UI.Blazor/Pages/EmbeddedChatPage.razor
```diff
@@ -0,0 +1,114 @@
...
+            var newChatFooter = await GetFooterModel(newChat!, account, cancellationToken);
+            return new (newChat!, account, newChatFooter, screenSize);
+        }
+        catch (Exception e) when (e is not OperationCanceledException) {
+            Log.LogWarning(e, "Failed to get Chat #{ChatId}", chatId);
+            return new ChatPage.Model(SpecialChat.Unavailable, account, new(), screenSize);
+        }
+    }
+
+    protected override bool ShouldRender()
```
crui3er, line 110: Looks like the same as on a base class

src/dotnet/Chat.UI.Blazor/Pages/EmbeddedChatPage.razor
```diff
@@ -0,0 +1,114 @@
+@page "/embedded/{ChatSid}"
+@inherits ChatPage
```
crui3er, line 2: Is it not better to introduce ChatPageBase class?
Now it seems a lot of significant logic from ChatPage class is overriden: markup and ComputeState.
And it makes harder to follow what in fact is inherited and what is replaced.

src/dotnet/UI.Blazor.App/Pages/Test/EmbeddedTestPage.razor
```diff
@@ -0,0 +1,45 @@
+@page "/test/embedded"
+@using ActualChat.Hosting
+
+@{
+    var baseUri = new Uri(HostInfo.BaseUrl);
+    var domain = HostInfo.IsDevelopmentInstance
```
crui3er, line 6: Can you explain with comments what this trick with "cdn.local.actual.chat" domain means?

.config/local.actual.chat/nginx/nginx.conf
```diff
@@ -53,11 +53,28 @@ server {
...
+        proxy_read_timeout 1d;
+        proxy_connect_timeout 4;
+        proxy_send_timeout 1d;
+
+        include nginx.proxy-headers.conf;
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection $http_connection;
+    }
+
+    location /media/ {
```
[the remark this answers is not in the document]
alexis-kochetov: They are totally unrelated. I just need some working url different to the local.actual.chat to ensure that we are fine with CORS. So I reuse cdn.local.actual.chat

src/dotnet/Chat.Service/Chats.cs
```diff
@@ -304,6 +308,140 @@ async Task Remove(ChatEntryId entryId1)
...
+        // clone template chat
+        var cloneCommand = new IChatsBackend.ChangeCommand(
+            ChatId.None,
+            null,
+            new Change<ChatDiff> {
+                Create = new ChatDiff {
+                    Title = templateChat.Title,
+                    MediaId = templateChat.MediaId,
+                    Kind = ChatKind.Group,
+                    IsPublic = true,
```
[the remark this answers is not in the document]
alexis-kochetov: I don't have other scenarios - but in the future we can change this approach.

src/dotnet/Chat.Service/ChatsBackend.cs
alexis-kochetov: Don't see real reason to do so, just copied existing code. I have removed all cases with .AsQueryable()

src/dotnet/Chat.Service/Chats.cs
alexis-kochetov: It's not safe to allow join to anyone, so we use backend command to add guest author explicitly to the new chat (create for this guest user)

PR #1256, opened 2023-06-08
ios release build

alexis-kochetov:


src/dotnet/App.Maui/MauiDiagnostics.cs
```diff
@@ -41,9 +41,8 @@ public static IServiceCollection AddMauiDiagnostics(this IServiceCollection serv
     {
         services.AddSingleton<Disposer>();
         services.AddTracer(Tracer); // We don't want to have scoped tracers in MAUI app
-        services.AddSingleton(LoggerFactory);
-        services.Add(ServiceDescriptor.Singleton(typeof(ILogger<>), typeof(Logger<>)));
-        /*
+        // services.AddSingleton(LoggerFactory);
```
crui3er, line 44: Why is that commented out?

src/dotnet/App.Maui/MauiDiagnostics.cs
```diff
@@ -62,7 +60,7 @@ public static IServiceCollection AddMauiDiagnostics(this IServiceCollection serv
     private static LoggerConfiguration CreateSerilogLoggerConfiguration()
     {
         var configuration = new LoggerConfiguration()
-            .MinimumLevel.Is(LogEventLevel.Information)
+            .MinimumLevel.Is(LogEventLevel.Debug)
```
crui3er, line 63: I think better to return back min log level to Info

src/dotnet/App.Server/App.Server.csproj
```diff
@@ -10,6 +10,8 @@
     <InvariantGlobalization>true</InvariantGlobalization>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(RunConfiguration)' == 'Blazor Debug' " />
```
crui3er, line 13: Looks like these groups do nothing

src/dotnet/Directory.Build.props
```diff
@@ -110,58 +110,5 @@
     <Watch Remove="**\*.js" />
     <Watch Remove="**\*.css" />
   </ItemGroup>
-
-  <!-- Extra trimming roots -->
-  <ItemGroup>
```
crui3er, line 115: Won't it affect Android build?
Or we use their partial trimming and all assebmlies are considered to not be trimmed if it's not specified explicitly?

alexis-kochetov: Will test tomorrow, but we use partial trimming and our assemblies including stl are not marked with trim attribute explicitly

src/dotnet/App.Maui/MauiDiagnostics.cs
alexis-kochetov: @alexyakunin i have commented out this code because it doesn't work, and i don't understand why

alexyakunin: @crui3er maybe you can check what's going on there? Or it's just on iOS? I would test it on Windows app first to find out if it's an iOS-specific issue.

src/dotnet/App.Maui/MauiDiagnostics.cs
alexyakunin: Yep.

src/dotnet/App.Server/App.Server.csproj
```diff
@@ -10,6 +10,8 @@
...
+  <PropertyGroup Condition=" '$(RunConfiguration)' == 'Debug' " />
```
alexyakunin, line 14: Same here.

src/dotnet/Chat/Markup/Internal/ParserExt.cs
```diff
@@ -11,23 +11,23 @@ internal static class ParserExt
 
     public static Action<string>? DebugOutput { get; set; }
 
-    public static Parser<char, T> Debug<T>(this Parser<char, T> parser, Func<T, string> formatter)
+    public static Parser<char, T> Debug<T>(this Parser<char, T> parser, Func<T, string> formatter) where T: Markup
```
alexyakunin, line 14: I would rather remove "where" or put `[Conditional("DEBUG")]` here.

src/dotnet/Chat/Markup/Internal/ParserExt.cs
```diff
@@ -11,23 +11,23 @@ internal static class ParserExt
...
         => Constants.DebugMode.MarkupParser
             ? parser.Select(x => {
                 DebugOutput?.Invoke(formatter(x));
                 return x;
             })
             : parser;
-    public static Parser<char, T> Debug<T>(this Parser<char, T> parser, string title)
+    public static Parser<char, T> Debug<T>(this Parser<char, T> parser, string title) where T: Markup
```
alexyakunin, line 21: Same here.

src/dotnet/Chat/Markup/Internal/ParserExt.cs
alexyakunin: Another alternative is to use `where T : class` - i.e. no reason to limit this just to `Markup`.

src/dotnet/Chat/Markup/Internal/ParserExt.cs
```diff
@@ -52,10 +52,10 @@ internal static class ParserExt
 
     public static readonly Parser<char, Unit> Nothing = new NothingParser();
 
-    public static Parser<char, T> SafeTryOneOf<T>(params Parser<char, T>[] parsers)
+    public static Parser<char, T> SafeTryOneOf<T>(params Parser<char, T>[] parsers) where T: Markup
```
alexyakunin, line 55: Same here: `where T : class`.

alexyakunin (changes requested):

src/dotnet/App.Maui/MauiDiagnostics.cs
alexis-kochetov: Guys, we manage min level at MSFT logger config 
image

alexis-kochetov: So if we put min level Info there - we will not be able to log Debug

src/dotnet/App.Maui/MauiDiagnostics.cs
crui3er: > @alexyakunin i have commented out this code because it doesn't work, and i don't understand why

What does it mean 'doesn't work'? On Android and Window logging does work.

PR #1260, opened 2023-06-18
feat: add native Apple auth to iOS

evgeniy-ac:


src/dotnet/Users.Service/Controllers/MobileAuthController.cs
```diff
@@ -272,4 +322,26 @@ private async Task WriteAutoClosingMessage(CancellationToken cancellationToken)
...
     }
+
+    private async Task AuthenticateSessionWithPrincipal(
+        string sessionId,
+        ClaimsPrincipal principal,
+        CancellationToken cancellationToken)
+    {
+        var oldUser = HttpContext.User;
+        HttpContext.User = principal;
+        try
```
crui3er, line 333: Backet should be the same line as try

src/dotnet/Users.Service/Controllers/MobileAuthController.cs
```diff
@@ -272,4 +322,26 @@ private async Task WriteAutoClosingMessage(CancellationToken cancellationToken)
...
+        {
+            var helper = Services.GetRequiredService<ServerAuthHelper>();
+            await helper.UpdateAuthState(
+                    new Session(sessionId),
+                    HttpContext,
+                    cancellationToken)
+                .ConfigureAwait(false);
+        }
+        finally
+        {
```
crui3er, line 343: Backet should be the same line as finally

src/dotnet/Users.Service/Controllers/MobileAuthController.cs
```diff
@@ -153,6 +157,66 @@ public async Task SignIn(string sessionId, string scheme, CancellationToken canc
...
+        var identity = new ClaimsIdentity(options.ClaimsIssuer);
+
+        if (!email.IsNullOrEmpty()) {
+            identity.AddClaim(new Claim(ClaimTypes.Email, email));
+        }
+
+        if (!name.IsNullOrEmpty()) {
+            var names = name.Split(' ');
+            switch (names.Length)
+            {
```
crui3er, line 199: Backet should be the same line as switch

src/dotnet/Users.Service/Controllers/MobileAuthController.cs
```diff
@@ -153,6 +157,66 @@ public async Task SignIn(string sessionId, string scheme, CancellationToken canc
...
+        using var tokenResponse = await ExchangeCodeAsync(code, options, cancellationToken).ConfigureAwait(false);
+        if (tokenResponse.Error != null) {
+            Logger.LogError(tokenResponse.Error, $"{nameof(SignInAppleWithCode)} error.");
+
+            return BadRequest();
+        }
+
+        var identity = new ClaimsIdentity(options.ClaimsIssuer);
+
+        if (!email.IsNullOrEmpty()) {
```
crui3er, line 192: I see that you take email and name from the client. 
For Google I take them from `tokenResponse`. Is it not possible to do it here the same way?

crui3er (changes requested):

src/dotnet/Users.Service/Controllers/MobileAuthController.cs
evgeniy-ac: It's the only place where we can get this data
Also it's available only at the very first request

src/dotnet/Users.Service/Controllers/MobileAuthController.cs
evgeniy-ac: fixes

src/dotnet/Users.Service/Controllers/MobileAuthController.cs
evgeniy-ac: fixed

src/dotnet/Users.Service/Controllers/MobileAuthController.cs
evgeniy-ac: fixed

PR #1326, opened 2023-07-21
feat: Share links

crui3er:
Adds 'Share' item to Chat props, My Account settings tile, Avatar cards.
image
image
image
image
image
image
image
image
image
image


alexyakunin:
Chat Settings -> Public join link - do we need it? Feels excessive, at least on mobile. URLs aren't visible there anyway.

alexyakunin:
Maybe it even makes sense to remove it everywhere, assuming "Share" option stays & works as "Copy to clipboard" (w/ hint afterward). Thoughts?

alexyakunin:
"Who would you like to share it with?"
// "To" and "?" is added

alexyakunin:
"Share chat" - I would similarly remove the link there.

alexyakunin:
"Share chat" modal:
- "Share" button below - I would rather write "Share externally" & turn it into a link-style button, coz otherwise it messes up "Share" action.

Overall, I'm for this setup:

Desktop:
```
Share chat
[Share externally]

[Search box]
[Chat & ppl list]
-----------------
[Cancel] [Share]
```

Mobile:

```
Share chat // No "Share" in header
[Share externally]

[Search box]
[Chat & ppl list]
-----------------
[Cancel] [Share]
```

crui3er:
> Chat Settings -> Public join link - do we need it? Feels excessive, at least on mobile. URLs aren't visible there anyway.

removed

PR #1331, opened 2023-07-26
a lot of css refactoring, win app qr and link added, landing tile signin added, etc

iqmulator:
image
Header button (hover)
image
Header button inverted
image
Header button active bg for mobile
image
Tailwind class for chat, modals, panels header


iqmulator:
image
Remove part of css from razor

iqmulator:
image
Win app QR and web signin added

PR #1372, opened 2023-08-24
md:820px, update left and right chat panels

iqmulator:


src/dotnet/Chat.UI.Blazor/Components/LeftPanel/left-panel.css
```diff
@@ -4,13 +4,19 @@
     @apply bg-04;
 }
 .left-panel-buttons {
-    @apply flex md:w-30 h-full flex-x bg-04 z-10 px-1 md:px-3
+    @apply z-10;
```
frolyo, line 7: don't we use custom z-indices?

frolyo (approved):

PR #1407, opened 2023-08-30
troubleshoot recording guide

alexis-kochetov:
image

### DIFF
--- a/src/dotnet/Chat.UI.Blazor/Components/ChatAudioPanel/RecorderToggle.razor
+++ b/src/dotnet/Chat.UI.Blazor/Components/ChatAudioPanel/RecorderToggle.razor
@@ -97,6 +97,10 @@
         base.OnInitialized();
     }
 
+    protected override async Task OnInitializedAsync() {
+        await AudioRecorder.MicrophonePermission.Check(CancellationToken.None);
+    }
+
     protected override async Task OnAfterRenderAsync(bool firstRender) {
         if (!firstRender)
             return;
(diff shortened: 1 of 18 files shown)


src/dotnet/Chat.UI.Blazor/Components/ChatAudioPanel/RecorderToggle.razor
```diff
@@ -154,17 +178,42 @@
...
+            var state = State.LastNonErrorValue;
+            if (!state.StartRecordingAt.HasValue) {
+                _isTroubleshootGuideTriggered = false;
+                return;
+            }
+
+            var model = new GuideModal.Model(false, GuideType.Unknown);
+            var modalRef = await ModalUI.Show(model);
+            try {
+                await modalRef.WhenClosed.WaitAsync(CancellationToken.None);
```
crui3er, line 191: Why do you need this constuction with `.WaitAsync(CancellationToken.None);` and exception handling?

src/dotnet/Chat.UI.Blazor/Components/ChatAudioPanel/RecorderToggle.razor
```diff
@@ -86,6 +97,10 @@
         base.OnInitialized();
     }
 
+    protected override async Task OnInitializedAsync() {
```
crui3er: use arrow function

alexyakunin:
Fixed text:

It appears that Actual Chat is unable to start audio recording. The most likely cause is that the microphone access permission isn't granted, although it could also be due to an issue on our end.

To address this problem:

- [Click here] to perform a hot restart of the app.
- If that doesn't help, try closing and reopening the app.
- If the above steps are not effective, please verify whether the app has microphone access permission.

PR #1408, opened 2023-08-31
feat: bulk message deletion

evgeniy-ac:


src/dotnet/Chat.Service/Chats.cs
```diff
@@ -522,4 +530,78 @@ public virtual async Task<Unit> OnForwardTextEntries(Chats_ForwardTextEntries co
...
+            if (entry1 == null || !entry1.IsRemoved)
+                return;
+
+            entry1 = entry1 with { IsRemoved = false };
+            var upsertCommand = new ChatsBackend_UpsertEntry(entry1);
+            await Commander.Call(upsertCommand, true, cancellationToken).ConfigureAwait(false);
+        }
+
+        async ValueTask<ChatEntry?> GetRemovedEntry(ChatEntryId entryId)
+        {
```
crui3er: put brackets on the same line as method declaration

crui3er (approved):

src/dotnet/Chat.Service/Chats.cs
evgeniy-ac: done

PR #1470, opened 2023-09-18
Notes chat

alexis-kochetov:


src/dotnet/Chat.Service/AuthorsBackend.cs
```diff
@@ -203,6 +203,14 @@ public virtual async Task<AuthorFull> OnUpsert(AuthorsBackend_Upsert command, Ca
             // Create author, + we know here it's not a peer chat
             if (userId.IsNone)
                 throw new ArgumentOutOfRangeException(nameof(command), "UserId is required to create a new author.");
+            var chat = await ChatsBackend.Get(chatId, cancellationToken).ConfigureAwait(false);
+            if (chat == null || chat.AllowSingleAuthorOnly) {
```
crui3er: Why do you do a check whether author exists or no when the chat is null?

src/dotnet/Chat.Service/ChatsBackend.cs
```diff
@@ -708,6 +714,43 @@ await dbContext.ChatEntries
...
+            null,
+            new Change<ChatDiff> {
+                Create = new ChatDiff {
+                    Title = "Notes",
+                    Kind = ChatKind.Group,
+                    IsPublic = false,
+                    IsTemplate = false,
+                    AllowGuestAuthors = false,
+                    AllowAnonymousAuthors = false,
+                    AllowSingleAuthorOnly = true,
```
crui3er: Do we really need to persist AllowSingleAuthorOnly property? May be we can rely on Tag? Certain tags will indicate that it's a SingleAuthorOnly chat.

src/dotnet/Chat.Service/ChatsBackend.cs
```diff
@@ -708,6 +714,43 @@ await dbContext.ChatEntries
...
+                    Tag = Constants.Chat.Tags.Notes,
+                },
+            },
+            userId);
+        await Commander.Call(createNotesCommand, cancellationToken).ConfigureAwait(false);
+        // var author = await AuthorsBackend.EnsureJoined(chat.Id, userId, cancellationToken).ConfigureAwait(false);
```
crui3er: remove code in comments

src/dotnet/Chat.Service/AuthorsBackend.cs
alexis-kochetov: Because the chat has not been committed yet, so Get methods returns null

src/dotnet/Chat.Service/ChatsBackend.cs
alexis-kochetov: As we already use flags to evaluate permissions it makes sense to use same approach for new restriction

frolyo (approved):

src/dotnet/Chat.Service/ChatsBackend.cs
alexyakunin: IMO it's not a good idea to do it this way too.

Here is what I would do:
1. Rename `Tag` field to `SystemTag` (`Symbol`)
2. Rename `AllowSingleAuthorOnly` property to `HasSingleAuthor()`, which uses `SystemTag` to compute its value.

src/dotnet/Chat.Service/ChatsBackend.cs
```diff
@@ -718,6 +772,9 @@ public virtual async Task OnNewUserEvent(NewUserEvent eventCommand, Cancellation
 
         await JoinAnnouncementsChat(eventCommand.UserId, cancellationToken).ConfigureAwait(false);
 
+        var createNotesCommand = new ChatsBackend_CreateNotesChat(eventCommand.UserId);
```
alexyakunin: Let's extract this to method + use `Run` with `isolate = true`.

src/dotnet/Chat.Service/ChatsBackend.cs
```diff
@@ -752,6 +809,10 @@ public virtual async Task OnAuthorChangedEvent(AuthorChangedEvent eventCommand,
         if ((chat?.TemplatedForUserId.HasValue ?? false) && ownerRole != null && author.RoleIds.Contains(ownerRole.Id))
             return;
 
+        // and chats with predefined tags
+        if (!(chat?.Tag.IsNullOrEmpty() ?? true))
```
alexyakunin: That's one of reasons to use `Symbol` instead - no `null`, easy comparison.

src/dotnet/Chat.Service/Db/DbChat.cs
```diff
@@ -30,6 +30,8 @@ public class DbChat : IHasId<string>, IHasVersion<long>, IRequirementTarget
     public bool IsPublic { get; set; }
     public bool AllowGuestAuthors { get; set; }
     public bool AllowAnonymousAuthors { get; set; }
+    public bool AllowSingleAuthorOnly { get; set; }
```
alexyakunin: Kinda useless field if there going to be `SystemTag`.

src/dotnet/Chat.Service/Module/ChatDbInitializer.cs
```diff
@@ -128,6 +129,39 @@ private async Task EnsureFeedbackTemplateChatExists(CancellationToken cancellati
...
+        await using var _ = dbContext.ConfigureAwait(false);
+
+        var hasNotesChat = await dbContext.Chats
+            .AnyAsync(c => c.Tag == Constants.Chat.Tags.Notes, cancellationToken)
+            .ConfigureAwait(false);
+        if (hasNotesChat)
+            return;
+
+        try {
+            Log.LogInformation("There is no 'Notes' chat, creating one for each user");
```
alexyakunin: "each" -> "every".

tests/Chat.IntegrationTests/ChatOperationsTest.cs
```diff
@@ -70,6 +70,69 @@ public async Task CreateNewChat(bool isPublicChat)
...
+        joined.Permissions.Has(ChatPermissions.Invite).Should().BeTrue();
+
+        chat.Should().NotBeNull();
+
+        var rules = await chatsBackend.GetRules(chat.Id, new PrincipalId(account.Id, AssumeValid.Option), default);
+        rules.CanRead().Should().BeTrue();
+        rules.CanWrite().Should().BeTrue();
+        rules.CanJoin().Should().BeTrue();
+        rules.CanInvite().Should().BeTrue();
+        rules.CanEditProperties().Should().BeTrue();
```
alexyakunin, line 129: Btw, it makes sense to add these checks + rules generation for Notes chat - and it shouldn't be possible to edit anything there, as well as to delete this chat.

alexyakunin (changes requested):

PR #1564, opened 2023-10-18
feat: render virtual list with tiles to reduce re-rendering on new entries

alexis-kochetov:


src/dotnet/Chat.UI.Blazor/Services/ChatUI.cs
```diff
@@ -300,6 +303,14 @@ public async ValueTask<SyncedStateLease<ReadPosition>> LeaseReadPositionState(Ch
         return result;
     }
 
+    public async ValueTask<ComputedStateLease<Range<long>>> LeaseChatIdRangeState(ChatId chatId, CancellationToken cancellationToken)
+    {
+        var lease = await _chatIdRangeStates.Rent(chatId, cancellationToken).ConfigureAwait(false);
+        var result = new ComputedStateLease<Range<long>>(lease);
+        await result.WhenSynchronized(cancellationToken).ConfigureAwait(false);
```
alexyakunin, line 310: There is no timeout here.

src/dotnet/Chat.UI.Blazor/Services/ChatUI.cs
```diff
@@ -13,6 +13,7 @@ namespace ActualChat.Chat.UI.Blazor.Services;
 public partial class ChatUI : WorkerBase, IHasServices, IComputeService, INotifyInitialized
 {
     private readonly SharedResourcePool<Symbol, ISyncedState<ReadPosition>> _readPositionStates;
+    private readonly SharedResourcePool<Symbol, IComputedState<Range<long>>> _chatIdRangeStates;
```
alexyakunin, line 16: Why do we need this?

src/dotnet/Core/ComputedStateLease.cs
```diff
@@ -0,0 +1,122 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using ActualChat.Pooling;
+using Stl.Conversion;
+
+namespace ActualChat;
+
+public interface IComputedStateLease<T> : IComputedState<T>;
+
+public class ComputedStateLease<T, TKey, TState, TResource>(
```
alexyakunin, line 9: This is definitely a handy thing.

src/dotnet/Chat.UI.Blazor/Services/ChatUI.cs
```diff
@@ -212,7 +214,8 @@ public virtual async Task<Trimmed<int>> GetUnreadCount(ChatId chatId, Cancellati
     [ComputeMethod]
     public virtual async Task<bool> IsEmpty(ChatId chatId, CancellationToken cancellationToken)
     {
-        var idRange = await Chats.GetIdRange(Session, chatId, ChatEntryKind.Text, cancellationToken).ConfigureAwait(false);
+        using var idRangeLease = await LeaseChatIdRangeState(chatId, cancellationToken).ConfigureAwait(false);
```
alexyakunin, line 217: So this is the only usage - can you clarify why it's necessary here?

alexyakunin (changes requested):

src/dotnet/Chat.UI.Blazor/Services/ChatUI.cs
alexis-kochetov: I also use it at ChatView to avoid dependency on ChatIdRange when we are rendering messages far from the end

src/dotnet/Chat.UI.Blazor/Services/ChatUI.cs
alexis-kochetov: I don't get you, we should wait for the GetIdRange call, what do you want to return if we won't wait for the call completion?

PR #1602, opened 2023-10-24
Close #1600 fix: Win app has white screen after resuming after hibernation/sleep

### LINKED ISSUE #1600: Win app has white screen after resuming after hibernation/sleep

crui3er:


src/dotnet/UI.Blazor/Services/BrowserInit/browser-init.ts
```diff
@@ -63,6 +63,10 @@ export class BrowserInit {
         }
     }
 
+    public static isStateOk() {
```
frolyo: can be expression method

frolyo (approved):

src/dotnet/UI.Blazor/Services/BrowserInit/browser-init.ts
crui3er: Looks like we do not use such approach in TS

PR #1701, opened 2023-12-17
feat: email verification

evgeniy-ac:


Directory.Packages.props
```diff
@@ -216,4 +217,4 @@
     <PackageVersion Include="RestEase.SourceGenerator" Version="1.6.4" PrivateAssets="All" IncludeAssets="runtime; build; native; contentfiles; analyzers" />
     <PackageVersion Include="Stl.Generators" Version="$(StlFusionVersion)" PrivateAssets="All" IncludeAssets="runtime; build; native; contentfiles; analyzers" />
   </ItemGroup>
-</Project>
+</Project>
```
frolyo: accidental?

src/dotnet/Users.Service/Email/EmailSender.cs
```diff
@@ -0,0 +1,33 @@
+using ActualChat.Users.Module;
+using MailKit.Net.Smtp;
+using MimeKit;
+
+namespace ActualChat.Users.Email;
+
+public interface IEmailSender
```
frolyo, line 7: Afair we have SmsGateway, so maybe call then consistently - both Sender or both Gateway...

frolyo (approved):

alexis-kochetov (approved):

src/dotnet/Users.Service/Email/EmailSender.cs
evgeniy-ac: Actually you don't have `SmsGateway`. It's `ITextMessageGateway` and IMO it's too abstract and its name conflicts with a `Gateway` pattern. `SmsSender` is more clear.

Directory.Packages.props
evgeniy-ac: done

PR #1709, opened 2023-12-22
Feat/1xxx move chat2place

crui3er:
Make it possible to move chat to place for chat owners

image
image

### DIFF
--- a/src/dotnet/Chat.Service/ChatsMigration.cs
+++ b/src/dotnet/Chat.Service/ChatsMigration.cs
@@ -1,4 +1,5 @@
 using ActualChat.Contacts;
+using ActualChat.Users;
 
 namespace ActualChat.Chat;
 
@@ -48,6 +49,14 @@ public class ChatsMigration(IServiceProvider services) : IChatsMigration
             await Commander.Call(backCommand2, true, cancellationToken).ConfigureAwait(false);
             Log.LogInformation("Completed ContactsMigrationBackend_MoveChatToPlace");
             executed = true;
+        }
+
+        {
+            Log.LogInformation("About to perform UserChatSettingsMigrationBackend_MoveChatToPlace");
+            var backCommand3 = new UserChatSettingsMigrationBackend_MoveChatToPlace(chatId, placeId);
+            var hasChanges = await Commander.Call(backCommand3, true, cancellationToken).ConfigureAwait(false);
+            Log.LogInformation("Completed UserChatSettingsMigrationBackend_MoveChatToPlace");
+            executed |= hasChanges;
         }
 
         Log.LogInformation("Completed ChatsMigration_MoveToPlace");
--- a/src/dotnet/Chat.Service/ChatsMigration.cs
+++ b/src/dotnet/Chat.Service/ChatsMigration.cs
@@ -1,0 +1,56 @@
+using ActualChat.Contacts;
+
+namespace ActualChat.Chat;
+
+public class ChatsMigration(IServiceProvider services) : IChatsMigration
+{
+    private IChats Chats { get; } = services.GetRequiredService<IChats>();
+    private IPlaces Places { get; } = services.GetRequiredService<IPlaces>();
+    private IContacts Contacts { get; } = services.GetRequiredService<IContacts>();
+    private ICommander Commander { get; } = services.GetRequiredService<ICommander>();
+    private ILogger Log { get; } = services.LogFor<ChatsMigration>();
+
+    // [CommandHandler]
+    public virtual async Task<bool> OnMoveToPlace(ChatsMigration_MoveToPlace command, CancellationToken cancellationToken)
+    {
+        if (Computed.IsInvalidating())
+            return default; // It just spawns other commands, so nothing to do here
+
+        var (session, chatId, placeId) = command;
+        var executed = false;
+        Log.LogInformation("About to perform ChatsMigration_MoveToPlace");
+        var chat = await Chats.Get(session, chatId, cancellationToken).ConfigureAwait(false);
+        Log.LogInformation("Chat for chat id '{ChatId}' is {Chat}", chatId, chat);
+        if (chat != null) {
+            if (chat.Id.Kind != ChatKind.Group)
+                throw StandardError.Constraint("You may perform 'move to place' operation only on a group chat.");
+            if (!chat.Rules.IsOwner())
+                throw StandardError.Constraint("You should be a chat owner to perform 'move to place' operation.");
+
+            var place = await Places.Get(session, placeId, cancellationToken).Require().ConfigureAwait(false);
+            if (!place.Rules.IsOwner())
+                throw StandardError.Constraint("You should be a place owner to perform 'move to place' operation.");
+
+            Log.LogInformation("About to perform ChatsMigrationBackend_MoveToPlace");
+            var backCommand = new ChatsMigrationBackend_MoveToPlace(chatId, placeId);
+            await Commander.Call(backCommand, true, cancellationToken).ConfigureAwait(false);
+            Log.LogInformation("Completed ChatsMigrationBackend_MoveToPlace");
+            executed = true;
+        }
+
+        var placeChatId = new PlaceChatId(PlaceChatId.Format(placeId, chatId.Id));
+        var newChatId = (ChatId)placeChatId;
+        var contact = await Contacts.GetForChat(session, newChatId, cancellationToken).ConfigureAwait(false);
+        Log.LogInformation("Contact for chat id '{ChatId}' is {Contact}", newChatId, contact);
+        if (contact == null || contact.PlaceId.IsNone) {
+            Log.LogInformation("About to perform ContactsMigrationBackend_MoveChatToPlace");
+            var backCommand2 = new ContactsMigrationBackend_MoveChatToPlace(chatId, placeId);
+            await Commander.Call(backCommand2, true, cancellationToken).ConfigureAwait(false);
+            Log.LogInformation("Completed ContactsMigrationBackend_MoveChatToPlace");
+            executed = true;
+        }
+
+        Log.LogInformation("Completed ChatsMigration_MoveToPlace");
+        return executed;
+    }
+}
--- a/src/dotnet/Chat.Service/ChatsMigration.cs
+++ b/src/dotnet/Chat.Service/ChatsMigration.cs
@@ -42,7 +42,7 @@ public class ChatsMigration(IServiceProvider services) : IChatsMigration
         var newChatId = (ChatId)placeChatId;
         var contact = await Contacts.GetForChat(session, newChatId, cancellationToken).ConfigureAwait(false);
         Log.LogInformation("Contact for chat id '{ChatId}' is {Contact}", newChatId, contact);
-        if (contact == null || contact.PlaceId.IsNone) {
+        if (contact == null || contact.PlaceId.IsNone || !contact.IsStored()) {
             Log.LogInformation("About to perform ContactsMigrationBackend_MoveChatToPlace");
             var backCommand2 = new ContactsMigrationBackend_MoveChatToPlace(chatId, placeId);
             await Commander.Call(backCommand2, true, cancellationToken).ConfigureAwait(false);
--- a/src/dotnet/Chat.Service/ChatsMigrationBackend.cs
+++ b/src/dotnet/Chat.Service/ChatsMigrationBackend.cs
@@ -1,0 +1,426 @@
+using ActualChat.Chat.Db;
+using Microsoft.EntityFrameworkCore;
+using Stl.Fusion.EntityFramework;
+
+namespace ActualChat.Chat;
+
+public class ChatsMigrationBackend(IServiceProvider services): DbServiceBase<ChatDbContext>(services), IChatsMigrationBackend
+{
+    private IAuthorsBackend AuthorsBackend { get; } = services.GetRequiredService<IAuthorsBackend>();
+    private IChatsBackend ChatsBackend { get; } = services.GetRequiredService<IChatsBackend>();
+
+    public virtual async Task OnMoveToPlace(ChatsMigrationBackend_MoveToPlace command, CancellationToken cancellationToken)
+    {
+        var (chatId, placeId) = command;
+
+        if (Computed.IsInvalidating()) {
+            _ = ChatsBackend.Get(chatId, default);
+            _ = ChatsBackend.GetPublicChatIdsFor(placeId, default);
+            return;
+        }
+
+        var chatSid = chatId.Value;
+        var placeChatId = new PlaceChatId(PlaceChatId.Format(placeId, chatId.Id));
+        var newChatId = (ChatId)placeChatId;
+        var newChatSid = newChatId.Value;
+
+        var migratedRoles = new List<MigratedRole>();
+        var migratedAuthors = new List<MigratedAuthor>();
+
+        Log.LogInformation("About to move chat '{ChatId}' to place '{PlaceId}'", chatSid, placeId);
+
+        var dbContext = await CreateCommandDbContext(cancellationToken).ConfigureAwait(false);
+        await using var __ = dbContext.ConfigureAwait(false);
+
+        await UpdateChat(dbContext, chatSid, newChatSid, cancellationToken).ConfigureAwait(false);
+
+        await UpdateRoles(dbContext, chatSid, newChatId, migratedRoles, cancellationToken).ConfigureAwait(false);
+
+        await UpdateAuthors(dbContext, chatSid, newChatId, migratedRoles, migratedAuthors, cancellationToken).ConfigureAwait(false);
+
+        await UpdateChatEntries(dbContext, chatSid, newChatId, migratedAuthors, cancellationToken).ConfigureAwait(false);
+
+        await UpdateTextEntryAttachments(dbContext, chatSid, newChatSid, cancellationToken).ConfigureAwait(false);
+
+        await UpdateMentions(dbContext, chatSid, newChatSid, migratedAuthors, cancellationToken).ConfigureAwait(false);
+
+        await UpdateReactions(dbContext, chatSid, newChatSid, migratedAuthors, cancellationToken).ConfigureAwait(false);
+
+        await UpdateReactionSummaries(dbContext, chatSid, newChatSid, migratedAuthors, cancellationToken).ConfigureAwait(false);
+    }
+
+    private async Task UpdateChat(ChatDbContext dbContext, string chatSid, string newChatSid,
+        CancellationToken cancellationToken)
+    {
+        var updateCount = await dbContext.Chats
+            .Where(c => c.Id == chatSid)
+            .ExecuteUpdateAsync(setters => setters.SetProperty(c => c.Id, c => newChatSid), cancellationToken)
+            .ConfigureAwait(false);
+        AssertOneUpdated(updateCount);
+
+        Log.LogInformation("Updated chat record");
+    }
+
+    private async Task UpdateRoles(
+        ChatDbContext dbContext,
+        string chatSid,
+        ChatId newChatId,
+        ICollection<MigratedRole> migratedRoles,
+        CancellationToken cancellationToken)
+    {
+        var dbRoles = await dbContext.Roles
+            .Where(c => c.ChatId == chatSid)
+            .OrderBy(c => c.LocalId)
+            .ToListAsync(cancellationToken)
+            .ConfigureAwait(false);
+
+        foreach (var dbRole in dbRoles) {
+            var originalRole = dbRole.ToModel();
+            var newRoleId = new RoleId(RoleId.Format(newChatId, dbRole.LocalId));
+
+            var updateCount = await dbContext.Roles
+                .Where(c => c.Id == dbRole.Id)
+                .ExecuteUpdateAsync(setters => setters
+                        .SetProperty(c => c.Id, c => newRoleId)
+                        .SetProperty(c => c.ChatId, c => newChatId),
+                    cancellationToken)
+                .ConfigureAwait(false);
+            AssertOneUpdated(updateCount);
+
+            var newRole = originalRole with { Id = newRoleId };
+            migratedRoles.Add(new MigratedRole(originalRole, newRole));
+        }
+
+        Log.LogInformation("Updated {Count} role records", dbRoles.Count);
+    }
+
+    private async Task UpdateAuthors(
+        ChatDbContext dbContext,
+        string chatSid,
+        ChatId newChatId,
+        IReadOnlyCollection<MigratedRole> migratedRoles,
+        ICollection<MigratedAuthor> migratedAuthors,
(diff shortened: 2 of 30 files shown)


src/dotnet/Chat.Service/ChatsMigration.cs
```diff
@@ -0,0 +1,65 @@
+using ActualChat.Contacts;
+using ActualChat.Users;
+
+namespace ActualChat.Chat;
+
+public class ChatsMigration(IServiceProvider services) : IChatsMigration
```
alexyakunin: Let's use either `ChatMigrator` or `ChatMigration`.

src/dotnet/Chat.Service/ChatsMigration.cs
```diff
@@ -0,0 +1,65 @@
...
+            return default; // It just spawns other commands, so nothing to do here
+
+        var (session, chatId, placeId) = command;
+        var executed = false;
+        Log.LogInformation("About to perform ChatsMigration_MoveToPlace");
+        var chat = await Chats.Get(session, chatId, cancellationToken).ConfigureAwait(false);
+        Log.LogInformation("Chat for chat id '{ChatId}' is {Chat}", chatId, chat);
+        if (chat != null) {
+            if (chat.Id.Kind != ChatKind.Group)
+                throw StandardError.Constraint("You may perform 'move to place' operation only on a group chat.");
```
alexyakunin: "Only group chats can be moved to a Place".

src/dotnet/Chat.Service/ChatsMigration.cs
```diff
@@ -0,0 +1,65 @@
...
+            if (!chat.Rules.IsOwner())
+                throw StandardError.Constraint("You should be a chat owner to perform 'move to place' operation.");
+
+            var place = await Places.Get(session, placeId, cancellationToken).Require().ConfigureAwait(false);
+            if (!place.Rules.IsOwner())
+                throw StandardError.Constraint("You should be a place owner to perform 'move to place' operation.");
```
alexyakunin: "You must be the Owner of this chat to perform the migration."

src/dotnet/Chat.Service/ChatsMigration.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            Log.LogInformation("About to perform ChatsMigrationBackend_MoveToPlace");
```
alexyakunin: I'd rather stick to "X: starting" and "X: completed" format.

alexyakunin: That's because we start most of our scope-like logging calls with the method / command name.

src/dotnet/Chat.Service/ChatsMigration.cs
```diff
@@ -0,0 +1,65 @@
...
+            var backCommand = new ChatsMigrationBackend_MoveToPlace(chatId, placeId);
```
alexyakunin: "backCommand" is kinda confusing, let's use "backendCmd" or "backendCommand"

src/dotnet/Chat.Service/ChatsMigration.cs
```diff
@@ -0,0 +1,65 @@
...
+            Log.LogInformation("Completed ChatsMigrationBackend_MoveToPlace");
+            executed = true;
+        }
+
+        var placeChatId = new PlaceChatId(PlaceChatId.Format(placeId, chatId.Id));
+        var newChatId = (ChatId)placeChatId;
+        var contact = await Contacts.GetForChat(session, newChatId, cancellationToken).ConfigureAwait(false);
+        Log.LogInformation("Contact for chat id '{ChatId}' is {Contact}", newChatId, contact);
+        if (contact == null || contact.PlaceId.IsNone || !contact.IsStored()) {
+            Log.LogInformation("About to perform ContactsMigrationBackend_MoveChatToPlace");
```
alexyakunin: Same here.

src/dotnet/Chat.Service/ChatsMigration.cs
```diff
@@ -0,0 +1,65 @@
...
+            var backCommand2 = new ContactsMigrationBackend_MoveChatToPlace(chatId, placeId);
+            await Commander.Call(backCommand2, true, cancellationToken).ConfigureAwait(false);
+            Log.LogInformation("Completed ContactsMigrationBackend_MoveChatToPlace");
+            executed = true;
+        }
+
+        {
+            Log.LogInformation("About to perform UserChatSettingsMigrationBackend_MoveChatToPlace");
```
alexyakunin: Same here.

src/dotnet/Chat.Service/ChatsMigration.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        var placeChatId = new PlaceChatId(PlaceChatId.Format(placeId, chatId.Id));
+        var newChatId = (ChatId)placeChatId;
+        var contact = await Contacts.GetForChat(session, newChatId, cancellationToken).ConfigureAwait(false);
+        Log.LogInformation("Contact for chat id '{ChatId}' is {Contact}", newChatId, contact);
+        if (contact == null || contact.PlaceId.IsNone || !contact.IsStored()) {
+            Log.LogInformation("About to perform ContactsMigrationBackend_MoveChatToPlace");
+            var backCommand2 = new ContactsMigrationBackend_MoveChatToPlace(chatId, placeId);
+            await Commander.Call(backCommand2, true, cancellationToken).ConfigureAwait(false);
+            Log.LogInformation("Completed ContactsMigrationBackend_MoveChatToPlace");
```
alexyakunin: Same here.

src/dotnet/Chat.Service/ChatsMigration.cs
```diff
@@ -0,0 +1,65 @@
...
+                throw StandardError.Constraint("You should be a chat owner to perform 'move to place' operation.");
+
+            var place = await Places.Get(session, placeId, cancellationToken).Require().ConfigureAwait(false);
+            if (!place.Rules.IsOwner())
+                throw StandardError.Constraint("You should be a place owner to perform 'move to place' operation.");
+
+            Log.LogInformation("About to perform ChatsMigrationBackend_MoveToPlace");
+            var backCommand = new ChatsMigrationBackend_MoveToPlace(chatId, placeId);
+            await Commander.Call(backCommand, true, cancellationToken).ConfigureAwait(false);
+            Log.LogInformation("Completed ChatsMigrationBackend_MoveToPlace");
```
alexyakunin: Same here.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+
+        var chatSid = chatId.Value;
+        var placeChatId = new PlaceChatId(PlaceChatId.Format(placeId, chatId.Id));
+        var newChatId = (ChatId)placeChatId;
+        var newChatSid = newChatId.Value;
+
+        var migratedRoles = new List<MigratedRole>();
+        var migratedAuthors = new List<MigratedAuthor>();
+
+        Log.LogInformation("About to move chat '{ChatId}' to place '{PlaceId}'", chatSid, placeId);
```
alexyakunin: Same here.

src/dotnet/Chat.Service/ChatsMigration.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            Log.LogInformation("About to perform UserChatSettingsMigrationBackend_MoveChatToPlace");
+            var backCommand3 = new UserChatSettingsMigrationBackend_MoveChatToPlace(chatId, placeId);
+            var hasChanges = await Commander.Call(backCommand3, true, cancellationToken).ConfigureAwait(false);
+            Log.LogInformation("Completed UserChatSettingsMigrationBackend_MoveChatToPlace");
+            executed |= hasChanges;
+        }
+
+        Log.LogInformation("Completed ChatsMigration_MoveToPlace");
+        return executed;
```
alexyakunin: IMO the result here should be `Unit` or `void`.

alexyakunin: I.e. we shouldn't provide a dual signal. What "false" exactly means here? I'd rather stick to an exception or nothing.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+            .Where(c => c.ChatId == chatSid)
+            .OrderBy(c => c.LocalId)
+            .ToListAsync(cancellationToken)
+            .ConfigureAwait(false);
+
+        foreach (var dbAuthor in dbAuthors) {
+            var originalAuthor = dbAuthor.ToModel();
+            var userId = originalAuthor.UserId;
+            if (userId.IsNone)
+                throw StandardError.NotSupported($"Migration for author with empty user id is not supported. AuthorId is '{dbAuthor.Id}'.");
```
alexyakunin: It should be `StandardError.Internal`, right? Normally this should never happen.

alexyakunin: "Can't proceed with the migration: found an author with no associated user."

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+            if (dbAuthor.IsAnonymous)
+                throw StandardError.NotSupported($"Migration for anonymous author is not supported. AuthorId is '{dbAuthor.Id}'.");
```
alexyakunin: "Can't proceed with the migration: anonymous author migration isn't supported yet."

Ideally we should handle this case though.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+            var authorSid = mentionSid.Substring(mentionIdAuthorPrefix.Length);
+            var newAuthorId = migratedAuthors.First(c => OrdinalEquals(c.OriginalAuthor.Id.Value, authorSid)).NewId;
+            var newMentionId = new MentionId(newAuthorId, AssumeValid.Option);
+            var newMentionSid = newMentionId.Value;
+#pragma warning disable CA1307
+            updateCount = await dbContext.Mentions
+                .Where(c => c.ChatId == chatSid)
+                .Where(c => c.MentionId == mentionSid)
+                .ExecuteUpdateAsync(setters => setters
+                        .SetProperty(c => c.Id,
```
alexyakunin: Wrong indent.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+        CancellationToken cancellationToken)
+    {
+        // TODO(DF): should we update MediaId and ThumbnailMediaId somehow?
+        var attachmentIdPrefix = chatSid + ":0:";
+        var attachmentNewIdPrefix = newChatSid + ":0:";
+        var attachmentNewIdPrefixLength = attachmentIdPrefix.Length;
+        var updateCount = await dbContext.TextEntryAttachments
+            .Where(c => c.Id.StartsWith(attachmentIdPrefix))
+            .ExecuteUpdateAsync(setters => setters
+                    .SetProperty(c => c.EntryId,
```
alexyakunin: Wrong double indent.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+        }
+
+        for (int i = 0; i < 2; i++) {
+            var entryKind = (ChatEntryKind)i;
+            var prefix = newChatId + ":" + i + ":";
+            updateCount = await dbContext.ChatEntries
+                .Where(c => c.ChatId == chatSid)
+                .Where(c => c.Kind == entryKind)
+                .ExecuteUpdateAsync(setters => setters
+                        .SetProperty(c => c.Id, c => prefix + c.LocalId)
```
alexyakunin: Wrong double indent.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+                        .SetProperty(c => c.ChatId, newChatId),
+                    cancellationToken)
+                .ConfigureAwait(false);
+
+            Log.LogInformation("Updated Id and ChatId for {Count} '{Kind}' chat entry records", updateCount, entryKind.ToString());
+        }
+
+        // NOTE: I expect that we have not many entries with forward fields filled in so far
```
alexyakunin: "we don't have"

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+        // hence it's ok to fetch them all at once.
+        var chatEntriesToUpdateForwardFields = await dbContext.ChatEntries
+            .Where(c => c.ForwardedAuthorId != null && c.ForwardedAuthorId.StartsWith(chatSid))
+            .ToListAsync(cancellationToken)
+            .ConfigureAwait(false);
+
+        updateCount = 0;
+        foreach (var dbChatEntry in chatEntriesToUpdateForwardFields) {
+            var forwardedAuthorId = dbChatEntry.ForwardedAuthorId.RequireNonEmpty("ForwardedAuthorId");
```
alexyakunin: "ForwardedAuthorId" -> nameof(dbChatEntry.ForwardedAuthorId)

alexyakunin: And the same further.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+    private async Task UpdateMentions(
+        ChatDbContext dbContext,
+        string chatSid,
+        string newChatSid,
+        IReadOnlyCollection<MigratedAuthor> migratedAuthors,
+        CancellationToken cancellationToken)
+    {
+        int updateCount;
+
+        var mentionSids = await dbContext.Mentions
```
alexyakunin: We need to update mentions in text as well...

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+
+    private async Task UpdateReactionSummaries(
+        ChatDbContext dbContext,
+        string chatSid,
+        string newChatSid,
+        IReadOnlyCollection<MigratedAuthor> migratedAuthors,
+        CancellationToken cancellationToken)
+    {
+        #pragma warning disable CA1307
+        var updateCount = await dbContext.ReactionSummaries
```
alexyakunin: Maybe makes sense to recompute these - i.e. delete the old ones & run a command (or commands) which will recompute the summaries.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+
+        await dbContext.SaveChangesAsync(cancellationToken).ConfigureAwait(false);
+
+        Log.LogInformation("Updated AuthorIds for {Count} reaction summary records", updateCount);
+    }
+
+    private static void AssertOneUpdated(int updateCount)
+        => AssertXUpdated(updateCount, 1);
+
+    private static void AssertXUpdated(int updateCount, int expectedUpdateCount)
```
alexyakunin: Rename to `AssertUpdated`; also, I'd rather use `N` instead of `X` in similar cases: `N` is more associated with a number, though `X` can be anything.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+    {
+        if (updateCount != expectedUpdateCount)
+            throw new InvalidOperationException($"Expected {expectedUpdateCount} row should be update, but {updateCount} rows were updated.");
```
alexyakunin: "Expected N rows to be updated, though only K rows were updated."

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    private static void AssertAtLeastOneUpdated(int updateCount)
+    {
+        if (updateCount < 1)
+            throw new InvalidOperationException("Expected at least 1 row should be update.");
```
alexyakunin: "Expected at least one row to be updated, though no rows were."

alexyakunin: It also feels like it's better to throw VersionMismatch or something like this here - i.e. any `ITransientError`.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
alexyakunin: It also feels like it's better to throw VersionMismatch or something like this here - i.e. any `ITransientError`.

src/dotnet/Contacts.Service/ContactsMigrationBackend.cs
```diff
@@ -0,0 +1,68 @@
...
+                .ConfigureAwait(false);
+            AssertOneUpdated(updateCount);
+        }
+
+        await dbContext.SaveChangesAsync(cancellationToken).ConfigureAwait(false);
+
+        // Place contacts should be added on adding authors to place root chat during processing ChatsMigrationBackend_MoveToPlace.
+    }
+
+    private static void AssertOneUpdated(int updateCount)
```
alexyakunin: I'd move all these methods to `ActualChat.Db` and implement them as extensions to `Task<int>` - `RequireUpdated`, `RequireOneUpdated`, `RequireAtLeastOneUpdated`.

alexyakunin (changes requested):
Left many minor comments, but the most important ones are:
- AssertXxxUpdated -> move to ActualChatDb + rename
- There seem to be no migration for mentions in Content.

src/dotnet/Chat.Service/ChatsMigration.cs
crui3er: done

src/dotnet/Chat.Service/ChatsMigration.cs
crui3er: done

src/dotnet/Chat.Service/ChatsMigration.cs
crui3er: done

src/dotnet/Chat.Service/ChatsMigration.cs
crui3er: done

src/dotnet/Chat.Service/ChatsMigration.cs
crui3er: done

src/dotnet/Chat.Service/ChatsMigration.cs
crui3er: done

src/dotnet/Chat.Service/ChatsMigration.cs
crui3er: done

src/dotnet/Chat.Service/ChatsMigration.cs
crui3er: done

src/dotnet/Chat.Service/ChatsMigration.cs
crui3er: done

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
crui3er: fixed

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
crui3er: Fixed error message. Will provide solution for migration later.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
crui3er: I think it's ok. Look to the entire statement. Also it's how rider reformat it's automatically.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
crui3er: I think it's ok. Look to the entire statement. Also it's how rider reformat it's automatically.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
crui3er: fixed

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
crui3er: fixed

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
crui3er: Why don't you like direct updates like it's implemented now?

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
crui3er: I think it's ok. Look to the entire statement. Also it's how rider reformat it's automatically.

src/dotnet/Contacts.Service/ContactsMigrationBackend.cs
crui3er: Done

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
crui3er: Fixed

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
crui3er: I would prefer not to use VersionMismatch or any other `ITransientError` since main purpose of this asserts was to ensure that written query is correct.

src/dotnet/Chat.Service/ChatsMigrationBackend.cs
crui3er: I would prefer not to use VersionMismatch or any other `ITransientError` since main purpose of this asserts was to ensure that written query is correct.

src/dotnet/Chat.Service/ChatsMigration.cs
crui3er: I return execution status because I want have indication that some job was executed. I assume that child commands may fail and I want to have possibility to reprocess.
Also I considered this functionality as to be used by admins only. But today you said that later we should provide it to every user, so may be we should discuss how to make it better in this case.

PR #1717, opened 2024-01-10
fix on render process gone crash anr

alexis-kochetov:


src/dotnet/App.Maui/Platforms/Android/AndroidWebChromeClient.cs
```diff
@@ -67,7 +67,8 @@ public AndroidWebChromeClient(WebChromeClient client, ComponentActivity activity
         _client = client;
         _fileChooser = fileChooser;
 
-        _requestPermissionLauncher = _activity.RegisterForActivityResult(
+        // ReSharper disable once NullCoalescingConditionIsAlwaysNotNullAccordingToAPIContract
+        _requestPermissionLauncher ??= _activity.RegisterForActivityResult(
```
crui3er, line 71: Why do you need `??=` operator here?

src/dotnet/App.Maui/WebView/MauiWebView.cs
```diff
@@ -131,7 +132,7 @@ public bool MarkDead()
 
     private void OnUnloaded(object? sender, EventArgs eventArgs)
     {
-#if !WINDOWS
+#if IOS
```
crui3er: I am sure that it's not safe for IOS neither.
Inside DisconnectHandler they do blocking await on main thread
https://github.com/dotnet/maui/blob/main/src/BlazorWebView/src/Maui/iOS/BlazorWebViewHandler.iOS.cs#L133
that causes deadlock when there are async disposable services which do not complete synchronously.

src/dotnet/App.Maui/Platforms/Android/AndroidWebViewClient.cs
```diff
@@ -31,13 +31,25 @@ public override bool OnRenderProcessGone(WebView? view, RenderProcessGoneDetail?
...
-        if (didCrash && markedDead) {
-            MainThread.BeginInvokeOnMainThread(() => MainPage.Current.RecreateWebView());
+            && ReferenceEquals(androidWebView, view)) {
+            if (mauiWebView.MarkDead()) {
+                androidWebView.ClearCache(false);
+                androidWebView.RemoveAllViews();
+                androidWebView.OnPause();
+                androidWebView.ClearHistory();
+                MainThread.BeginInvokeOnMainThread(() => {
+                    if (didCrash)
```
crui3er, line 45: Why do you need here different approaches to handle cases when render process has gone due to crash or not?

src/dotnet/App.Maui/Platforms/Android/AndroidWebChromeClient.cs
alexis-kochetov: When WebChromeClient is getting recreated we should keep single registration within the same activity - otherwise it will fail, e.g. we should register before the activity reach STARTED state

src/dotnet/App.Maui/WebView/MauiWebView.cs
alexis-kochetov: Then we should drop this case - I'm going to test iOS

src/dotnet/App.Maui/Platforms/Android/AndroidWebViewClient.cs
alexis-kochetov: Because it can be dropped due to memory pressure while the app is in background - and we will recreate it on getting foreground

PR #1724, opened 2024-01-17
feat: improve security for place chats

crui3er:
- adds checks that user have permissions to create a chat to specified place;
- adds checks that place member which is not owner should be able to create only private chats in a place;
- in ui chat type changing from private to public is blocked on 'New chat' modal for users who are not place owners;

image

### DIFF
--- a/src/dotnet/Chat.Service/Chats.cs
+++ b/src/dotnet/Chat.Service/Chats.cs
@@ -179,23 +179,22 @@ public class Chats(IServiceProvider services) : IChats
             : await Get(session, chatId, cancellationToken).ConfigureAwait(false);
 
         var changeCommand = new ChatsBackend_Change(chatId, expectedVersion, change.RequireValid());
-        if (change.Create.HasValue) {
+        if (change.IsCreate(out var chatDiff1)) {
             var account = await Accounts.GetOwn(session, cancellationToken).ConfigureAwait(false);
             account.Require(AccountFull.MustBeActive);
             changeCommand = changeCommand with {
                 OwnerId = account.Id,
             };
-            var chatDiff = changeCommand.Change.Create.Value;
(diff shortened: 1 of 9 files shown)


src/dotnet/Chat.Service/Chats.cs
```diff
@@ -181,18 +185,40 @@ public virtual async Task<Chat> OnChange(Chats_Change command, CancellationToken
             changeCommand = changeCommand with {
                 OwnerId = account.Id,
             };
+            var chatDiff = changeCommand.Change.Create.Value;
```
frolyo: there was `change.IsCreate(out var ...)` which you can use in if block above

alexis-kochetov (approved):

src/dotnet/Chat/ChatPermissionsExt.cs
```diff
@@ -35,10 +35,7 @@ public static void Require(this ChatPermissions available, ChatPermissions requi
     }
 
     public static Exception NotEnoughPermissions(ChatPermissions? required = null)
-    {
-        var message = "You can't perform this action: not enough permissions.";
-        return required.HasValue
-            ? new SecurityException($"{message} Requested permission: {required.Value}.")
-            : new SecurityException(message);
-    }
+        => required.HasValue
```
frolyo: what if we make StandardErrorEnoughPermission(string? message = null) instead? This will allow to get rid off ternary operator

frolyo (approved):

src/dotnet/Chat.Service/Chats.cs
crui3er: done

src/dotnet/Chat/ChatPermissionsExt.cs
crui3er: done

PR #1738, opened 2024-01-24
skeletons only pulse when visible

iqmulator:


src/dotnet/UI.Blazor/Components/Skeleton/chat-message-skeleton.lit.ts
```diff
@@ -42,4 +47,26 @@ class ChatMessageSkeletonLit extends LitElement {
...
+        });
+        this.observer.observe(this);
+    }
+
+    disconnectedCallback() {
+        super.disconnectedCallback();
+        this.observer.disconnect();
+    }
+
+    constructor() {
```
alexis-kochetov: Do we need an empty constructor? If still yes, let's move it to the top, right after fields

src/dotnet/UI.Blazor/Components/Skeleton/chat-message-skeleton.lit.ts
```diff
@@ -42,4 +47,26 @@ class ChatMessageSkeletonLit extends LitElement {
         let num = randomIntFromInterval(first, second);
         return MessageWidth[num];
     }
+
+    connectedCallback() {
+        super.connectedCallback();
+        const root = document.querySelector('.layout-body');
+        this.observer = new IntersectionObserver((entries, observer) => {
+            entries.forEach(entry => {
```
alexis-kochetov: It's better to use something like entries.some(e => e.isIntersecting)

alexis-kochetov (changes requested):

PR #1754, opened 2024-02-05
feat: NATS-based command queue

alexis-kochetov:


src/dotnet/Audio/WebMStreamConverter.cs
```diff
@@ -80,6 +80,7 @@ public WebMStreamConverter(MomentClockSet clocks, ILogger log)
                 throw;
             }
             finally {
+                readBuffer.Release();
```
alexyakunin: Seems unrelated?

src/dotnet/Core.Server/Commands/NatsCommandQueue.cs
```diff
@@ -0,0 +1,261 @@
...
+                Storage = StreamConfigStorage.File,
+                NumReplicas = queues.Settings.ReplicaCount,
+                Discard = StreamConfigDiscard.Old,
+                Retention = StreamConfigRetention.Workqueue,
+                AllowDirect = true,
+            };
+
+            return await js.CreateStreamAsync(config, lockCts.Token).ConfigureAwait(false);
+        }
+
```
alexyakunin: Empty line.

src/dotnet/Core/Commands/QueueId.cs
```diff
@@ -5,17 +5,17 @@ namespace ActualChat.Commands;
 [DataContract, MemoryPackable(GenerateType.VersionTolerant)]
 [StructLayout(LayoutKind.Auto)]
 public readonly partial record struct QueueId(
-    [property: DataMember(Order = 0), MemoryPackOrder(0)] int SharedKey,
+    [property: DataMember(Order = 0), MemoryPackOrder(0)] int ShardIndex,
     [property: DataMember(Order = 1), MemoryPackOrder(1)] QueuedCommandPriority Priority = QueuedCommandPriority.Normal)
 {
-    public QueueId WithShardKeyMask(int shardKeyMask)
-        => this with { SharedKey = SharedKey & shardKeyMask };
+    public QueueId WithShardIndexMask(int shardIndexMask)
```
alexyakunin: It's better to always use ShardScheme.ShardIndex here.

src/dotnet/Core/Commands/QueuedCommand.cs
```diff
@@ -1,45 +1,61 @@
...
-    [DataMember, MemoryPackOrder(4)] public Moment CreatedAt { get; init; }
-    [DataMember, MemoryPackOrder(5)] public Moment? StartedAt { get; init; }
-    [DataMember, MemoryPackOrder(6)] public Moment? CompletedAt { get; init; }
-    [DataMember, MemoryPackOrder(7)] public int TryIndex { get; init; }
-    [DataMember, MemoryPackOrder(8)] public string Error { get; init; } = "";
+    public static IMomentClock Clock { get; set; } = MomentClockSet.Default.CoarseSystemClock;
 
-    public static Symbol NewId()
-        => Ulid.NewUlid().ToString();
+    public Symbol Id { get; init; }
```
alexyakunin: I'd rather go with single `Ulid Id { get; }` here.

src/dotnet/Core/Commands/QueuedCommand.cs
```diff
@@ -1,45 +1,61 @@
...
+    protected QueuedCommand(Ulid ulid)
+    {
+        Ulid = ulid;
+        Id = Ulid.ToString();
+    }
+
     // Equality is based solely on Id property
-    public bool Equals(QueuedCommand? other) => other != null && Id.Equals(other.Id);
+    public virtual bool Equals(QueuedCommand? other) => other != null && Id.Equals(other.Id);
     public override int GetHashCode() => Id.GetHashCode();
```
alexyakunin, line 40: `Id.HashCode` is faster for `Symbol`.

alexyakunin:
A few more comments:
1. I'd rather refactor the setup for command queues. `QueueId` isn't necessary now, all you need is to reference `ShardScheme`.
2. It makes sense to move the whole code to `ActualChat.Server` - it isn't supposed to be used anywhere else.
3. `EnqueueOnCompletion` has to be removed: there is no warranty the code processing operation completion will actually run. But this item can be handled later, I guess.

A good way to handle #3 is to embed operation Id + DB it originates from into the queued command & make command processor to automatically take care of it (i.e. wait for op completion if the command has such an entry).

alexyakunin:
To clarify, `ShardScheme` is "a scheme for anything that can be sharded" - whether it's a backend sharding, DB sharding, or queue sharding. I recently renamed `ShardScheme.ByHostRole` -> `ShardScheme.ById` - in other words, shard schemes aren't necessarily bound to roles now, but `ServiceRoleBuilder` assumes that every host role it's used for has a shard scheme with the idetical `Id`.

src/dotnet/Core/Concurrency/InterlockedExt.cs
```diff
@@ -0,0 +1,16 @@
+namespace ActualChat.Concurrency;
+
+public static class InterlockedExt
+{
+    public static void ExchangeIfGreaterThan(ref long location, long value)
+    {
+        var current = Interlocked.Read(ref location);
+        while (current < value) {
```
frolyo, line 8: method name is `..IfGreaterThan` but in fact `IsLessThan`?

alexis-kochetov: Exchange if the value is greater than location - do you have suggestions to describe this logic?

src/dotnet/Core.Server/Commands/ShardCommandQueueScheduler.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            if (cancellationToken.IsCancellationRequested)
+                throw;
+        }
+        finally {
+            var commandCompletionTicks = Clock.UtcNow.Ticks;
+            InterlockedExt.ExchangeIfGreaterThan(ref _lastCommandTicks, commandCompletionTicks);
+        }
+    }
+
```
frolyo: redundant empty lines

src/dotnet/Core.Server/Commands/ShardEventQueueScheduler.cs
```diff
@@ -0,0 +1,129 @@
...
+                throw;
+        }
+        finally {
+            var commandCompletionTicks = Clock.UtcNow.Ticks;
+            InterlockedExt.ExchangeIfGreaterThan(ref _lastCommandTicks, commandCompletionTicks);
+
+            if (error == null)
+                await queueBackend.MarkCompleted(consumerPrefix, command, cancellationToken).ConfigureAwait(false);
+            context.TryComplete(cancellationToken);
+            await context.DisposeAsync().ConfigureAwait(false);
```
frolyo, line 126: Is it critical to call context.DisposeAsync()?
will this context dispose if cancellation is requested during finally block execution?

alexis-kochetov: Dispose doesn't require cancellation and moreover same approach we see at Commander.RunCommand\RunEvent - so let's keep it consistent

PR #1777, opened 2024-02-15
AI enabled search implementation

AlexUstinov:


src/dotnet/MLSearch/MLSearchChatId.cs
```diff
@@ -0,0 +1,22 @@
+using MemoryPack;
+
+namespace ActualChat.MLSearch;
+
+[DataContract, MemoryPackable(GenerateType.VersionTolerant)]
+public readonly partial struct MLSearchChatId : ISymbolIdentifier<MLSearchChatId>
```
alexyakunin, line 6: I'd rather use "context-unbound" composite identifiers here - i.e. something like SearchDocumentId, and make the IDs there parsable:  
- SearchDocumentKind Kind
- IsChatId(out ChatId chatId)
- IsXxxId(...)

alexis-kochetov (approved):

PR #1818, opened 2024-03-06
fix: use requestAnimationFrame for element manipulation at SideNav

alexis-kochetov:
to prevent layout thrashing

### DIFF
--- a/src/dotnet/UI.Blazor/Components/SideNav/side-nav.ts
+++ b/src/dotnet/UI.Blazor/Components/SideNav/side-nav.ts
@@ -277,8 +277,8 @@ class SideNavPullDetectGesture extends Gesture {
             }),
             DocumentEvents.capturedPassive.touchMove$.subscribe(e => move(e)),
             chatViewDiv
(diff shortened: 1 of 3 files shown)


alexyakunin (commented):
Does it help? Also, IMO one-time things like setting initial classes probably don't need RAF.

src/dotnet/UI.Blazor/Components/SideNav/side-nav.ts
```diff
@@ -266,8 +277,8 @@ class SideNavPullDetectGesture extends Gesture {
             }),
             DocumentEvents.capturedPassive.touchMove$.subscribe(e => move(e)),
             chatViewDiv
-                ? fromSubscription(fromEvent(chatViewDiv, 'scroll').subscribe(_ => this.dispose()))
-                : null,
+            ? fromSubscription(fromEvent(chatViewDiv, 'scroll').subscribe(_ => this.dispose()))
```
frolyo: why is indent removed?

frolyo (approved):

PR #1819, opened 2024-03-07
fix: can not see 'test pages' when select them in nav groups

crui3er:


src/dotnet/Chat.UI.Blazor/Components/Navbar/NavbarButtons.razor
```diff
@@ -99,6 +99,18 @@
     protected HostInfo HostInfo => Hub.HostInfo();
     protected Features Features => Hub.Features();
     protected ModalUI ModalUI => Hub.ModalUI;
+    protected NavbarUI NavbarUI => Hub.NavbarUI;
+    protected PanelsUI PanelsUI => Hub.PanelsUI;
+
+    protected override void OnInitialized() {
+        NavbarUI.SelectedGroupChanged += OnNavbarSelectedGroupChanged;
```
frolyo, line 106: we also have `<OnUIEvent .../>` to subscribe on ui events in razor

frolyo (approved):

src/dotnet/Chat.UI.Blazor/Components/Navbar/NavbarButtons.razor
crui3er: But this is not an UIEvent.
Changes in NavbarButtons is moving event handler code from NavbarGroupSelectionButton.
This is minor optimization.
Before we, in fact, had the same event handler to be executed for each button from NavbarButtons.

PR #1824, opened 2024-03-11
feat: audio stream buffering and restart on reconnect

alexis-kochetov:


src/dotnet/Streaming.Service/Services/Transcribers/GoogleTranscriber.cs
```diff
@@ -24,8 +24,8 @@ public partial class GoogleTranscriber : ITranscriber
     private static readonly Regex EndsWithWhitespaceOrEmptyRegex = EndsWithWhitespaceOrEmptyRegexFactory();
 
     private static readonly string RegionId = "us";
-    private static readonly TimeSpan SilentPrefixDuration = TimeSpan.FromSeconds(2);
-    private static readonly TimeSpan SilentSuffixDuration = TimeSpan.FromSeconds(4);
+    private static readonly TimeSpan SilentPrefixDuration = TimeSpan.FromSeconds(0);
```
crui3er, line 27: Is it OK that these intervals are zero now?

crui3er (approved):

src/dotnet/Streaming.Service/Services/Transcribers/GoogleTranscriber.cs
alexis-kochetov: Yep, we do not append any silence to out audio stream for Google transcriber.

PR #1833, opened 2024-03-12
feat: do not play same sounds too often

alexis-kochetov:


frolyo (approved):

src/dotnet/UI.Blazor/Services/TuneUI/sound-player.ts
```diff
@@ -3,13 +3,21 @@ import { audioContextSource } from '../../../Streaming.UI.Blazor/Services/audio-
 import { Log } from 'logging';
 
 const { debugLog, warnLog } = Log.get('SoundsPlayer');
+const PLAYBACK_COOLDOWN = 30 * 1000; // 30s
```
alexyakunin: IMO it should be configurable per sound effect.

alexyakunin (changes requested):

PR #1836, opened 2024-03-12
fix: place nav group is changed to chats nav group after peer chat is selected inside All group filter

crui3er:


frolyo (approved):

src/dotnet/Chat.UI.Blazor/Services/ChatUI.cs
```diff
@@ -392,8 +392,8 @@ private async Task SelectPlaceNavbarGroup(ChatId chatId)
             var listView = ChatListUI.ActiveChatListView.Value;
             if (listView != null && listView.PlaceId == SelectedPlaceId.Value) {
                 var settings = await listView.GetSettings().ConfigureAwait(false);
-                if (settings.Filter == ChatListFilter.People) {
-                    // Check if peer chat was shown for place people view
+                if (settings.Filter == ChatListFilter.People || settings.Filter == ChatListFilter.None) {
```
alexyakunin, line 398: IMO it makes sense to add comment here - I totally don't get what this means :)

alexyakunin (changes requested):

src/dotnet/Chat.UI.Blazor/Services/ChatUI.cs
crui3er: Added comments.

PR #1843, opened 2024-03-14
Parallel async indexing of different chats

AlexUstinov:


alexis-kochetov (approved):

src/dotnet/MLSearch.Service/Indexing/ChatIndexerWorker.cs
```diff
@@ -33,28 +73,27 @@ public async Task ExecuteAsync(int _shardIndex, CancellationToken cancellationTo
...
-                        Log.LogWarning("Event queue is full: We can't process till this indexing is fully complete.");
-                        while (!result.IsEndReached) {
-                            result = await dataIndexer.IndexNextAsync(e.Id, cancellationToken).ConfigureAwait(false);
-                        }
-                        Log.LogWarning("Event queue is full: exiting an element.");
-                    }
-                }
+
+        // TODO: add more logging and tracing
+        try {
```
andrew-kurochkin: With this changes the indexing flow will not be even. It will first index a single chat and than and only then try to move to the next one. I've intentionally made it so it will try to keep up with all chats. 
This is open for a discussion though.

src/dotnet/MLSearch.Service/Indexing/ChatIndexerWorker.cs
```diff
@@ -1,29 +1,69 @@
...
+            SingleReader = true,
             SingleWriter = false,
         });
+    private readonly ConcurrentDictionary<ChatId, Job?> _runningJobs = new(-1, ChannelCapacity);
+    private readonly SemaphoreSlim _semaphore = new(ChannelCapacity, ChannelCapacity);
 
-    public ChannelWriter<MLSearch_TriggerChatIndexing> Trigger => _channel.Writer;
+    public async ValueTask PostAsync(MLSearch_TriggerChatIndexing input, CancellationToken cancellationToken)
+    {
+        await _semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
```
andrew-kurochkin: I really-really don't want to acquire semaphore in one method and release it in another. It adds unnecessary complexity to manage and read

src/dotnet/MLSearch.Service/Indexing/ChatIndexerWorker.cs
```diff
@@ -33,28 +73,27 @@ public async Task ExecuteAsync(int _shardIndex, CancellationToken cancellationTo
...
             }
-            catch (Exception e){
-                Log.LogError(e.Message);
-                throw;
+            while (continueIndexing);
+        }
+        catch (OperationCanceledException) {
+            log.LogInformation("Indexing job for chat '{Id}' is cancelled.", input.Id);
+        }
+        catch (Exception e) {
```
andrew-kurochkin: I don't think eating up all exceptions is a good idea. 
In my opinion this and above should rethrow.

src/dotnet/MLSearch.Service/Indexing/ChatIndexerWorker.cs
```diff
@@ -1,29 +1,69 @@
...
     private const int ChannelCapacity = 10;
-    private ILogger? _log;
-    private ILogger Log => _log ??= loggerSource.GetLogger(GetType());
 
     private readonly Channel<MLSearch_TriggerChatIndexing> _channel =
-        Channel.CreateBounded<MLSearch_TriggerChatIndexing>(new BoundedChannelOptions(ChannelCapacity){
+        Channel.CreateBounded<MLSearch_TriggerChatIndexing>(new BoundedChannelOptions(ChannelCapacity) {
             FullMode = BoundedChannelFullMode.Wait,
-            SingleReader = false,
+            SingleReader = true,
```
andrew-kurochkin: I do believe this is a negative change, here is the reason.
With this change you're trying to put all triggered chats to index. This means in case of slow indexing (and it's actually a normal behavior) those chat identifiers will be buffered in the app memory vs in a queueing service. However the later is built to rescale at runtime, and, what is more important, it's persistent. With this change it will soak all the load into the memory of the app and in case of a failure all of those triggered events would disappear. Not only a single one, but all. Having a channel with a limited size buffer was already a compromise.

andrew-kurochkin (changes requested):
I do believe that we have to discuss those changes. I've commented on several aspects in comments, so you can read and understand my concerns. Also, I would like to call @alexyakunin into this discussion. Especially for the dictionary updates. I need another opinion here.

src/dotnet/MLSearch.Service/Indexing/ChatIndexerWorker.cs
AlexUstinov: This works in a separate logical thread, so other chats will be handled by other such threads in parallel.

src/dotnet/MLSearch.Service/Indexing/ChatIndexerWorker.cs
AlexUstinov: There is no other point to observe this exception, as there is no code waiting for this task completion.

src/dotnet/MLSearch.Service/Indexing/ChatIndexerWorker.cs
AlexUstinov: I introduced single dispatch flow to handle incoming events in this channel. This is why I switched to single reader.

src/dotnet/MLSearch.Service/Indexing/ChatIndexerWorker.cs
AlexUstinov: As you can see, we acquire semaphore by the dispatch flow but release it by worker flows. So it can't be done in the same method.

AlexUstinov:
After some investigation I decided to reopen it.
Code makes sense if my assumption that events from the same chat are routed to the same shard is valid.

andrew-kurochkin (approved):

PR #1898, opened 2024-04-03
Add publish copied chat functionality

crui3er:
- persists copied chats info in database;
- handles forwarded message click if source chat was copied to place and deleted afterwards;
- adds 'Copy chat' functionality to place settings modal (available for place owners who are admin users) 
image
image
- adds copied chat banner to chat view
image


src/dotnet/Api/Chat/CopiedChat.cs
```diff
@@ -0,0 +1,33 @@
+using ActualLab.Versioning;
+using MemoryPack;
+
+namespace ActualChat.Chat;
+
+[DataContract, MemoryPackable(GenerateType.VersionTolerant)]
+public partial record CopiedChat(
```
alexis-kochetov: ChatCopyState?

src/dotnet/Chat.Service/Chats.cs
```diff
@@ -680,24 +704,61 @@ public virtual async Task<Chat_CopyChatResult> OnCopyChat(Chat_CopyChat command,
...
+        var copiedChat = await GetCopiedChat(session, newChatId, cancellationToken).ConfigureAwait(false);
+
+        if (sourceChat != null && newChat.IsPublic != sourceChat.IsPublic) {
+            var changeChatCmd = new Chats_Change(session,
+                newChat.Id,
+                newChat.Version,
+                Change.Update(new ChatDiff {
+                    IsPublic = sourceChat.IsPublic,
+                }));
+            await Commander.Call(changeChatCmd, true, cancellationToken).ConfigureAwait(false);
```
alexis-kochetov: Why do you use `outermost` commands there instead of processing within same transaction?

alexis-kochetov (changes requested):

src/dotnet/Chat.Service/Chats.cs
crui3er: I think I can call ChatsBackend commands with IsOutermost = false, but ContactsBackend's command I have to call with IsOutermost = true since it uses another db context. Due to using ContactsBackend's command  I decided to run all child commands with IsOutermost = true.

src/dotnet/Api/Chat/CopiedChat.cs
crui3er: I can rename it. Should I also rename DbCopiedChat. I tried to avoid general used suffixes.

crui3er: Renamed

src/dotnet/Chat.Service/Chats.cs
crui3er: Fixed

alexis-kochetov (approved):

PR #1913, opened 2024-04-08
ChatIndexInitializer class unit tests

AlexUstinov:


tests/MLSearch.UnitTests/Indexing/Initializer/ChatIndexInitializerTests.cs
```diff
@@ -0,0 +1,214 @@
...
+
+namespace ActualChat.MLSearch.UnitTests.Indexing.Initializer;
+
+public class ChatIndexInitializerTests(ITestOutputHelper @out) : TestBase(@out)
+{
+    private const int InactiveShardIndex1 = 1;
+    private const int ActiveShardIndex = 2;
+    private const int InactiveShardIndex2 = 3;
+
+    private static readonly MethodInfo _onRun = typeof(ChatIndexInitializer).GetMethod(
```
crui3er, line 16: There is a new elegant way (since .NET8) to access non-public methods/fields UnsafeAccessorAttribute.
See https://github.com/Actual-Chat/actual-chat/blob/dev/src/dotnet/Core.Server/ServiceCollectionExt.cs#L23

crui3er (approved):

PR #1937, opened 2024-04-15
Switch to OpenSearch for contacts

alexis-kochetov:


src/dotnet/Search.Service/Module/SearchSettings.cs
```diff
@@ -3,12 +3,12 @@ namespace ActualChat.Search.Module;
 public sealed class SearchSettings
 {
     public bool IsSearchEnabled { get; set; }
-    public string ElasticCloudId { get; set; } = "";
-    public string ElasticApiKey { get; set; } = "";
-    public string ElasticLocalUri { get; set; } = "";
-    public TimeSpan ElasticRefreshInterval { get; set; } = TimeSpan.FromSeconds(30);
+    public string CloudId { get; set; } = "";
```
frolyo, line 6: Is it still actual?

frolyo (approved):

src/dotnet/Search.Service/Module/SearchSettings.cs
alexis-kochetov: I'm not sure - we have to deploy OpenSearch to the `dev` env first

AlexUstinov (approved):

PR #1953, opened 2024-04-17
Bugfix/1952 sometimes link preview is not generated

### LINKED ISSUE #1952: Sometimes link preview is not generated

frolyo:


src/dotnet/Media.Service/OpenGraphParser.cs
```diff
@@ -0,0 +1,41 @@
+using HtmlAgilityPack;
+
+namespace ActualChat.Media;
+
+public static class OpenGraphParser
+{
+    public static OpenGraph? Parse(string html)
+    {
+        var doc = new HtmlDocument();
```
alexis-kochetov, line 9: Aren't you afraid of parsing large documents here? Is there another way to limit `html` to some subset before the `</head>`

alexis-kochetov (approved):

src/dotnet/Media.Service/OpenGraphParser.cs
frolyo: Created issue #1956

PR #1962, opened 2024-04-19
update copy chat to place, fix file attachment and link preview

iqmulator:


alexis-kochetov (changes requested):
Let's add incomplete UI flag for chat copying

PR #1978, opened 2024-04-23
Feat/1971 delete place

### LINKED ISSUE #1971: Failed to remove place after chat was copied

crui3er:
feat: LocalFolderBlobStorage.Delete must throw exception if target object does not exist like it GoogleCloudBlobStorage.Delete does it.
fix: Delete place confirmation modal has 'Leave place' title.
feat: copying media on chat copying must make a copy for content.

### DIFF
--- a/src/dotnet/Chat.Service/ChatsBackend.CopyChat.cs
+++ b/src/dotnet/Chat.Service/ChatsBackend.CopyChat.cs
@@ -168,7 +168,15 @@ public partial class ChatsBackend
             .ConfigureAwait(false);
         // Make chat private, so only owner who executed copying can see the chat.
         // Publishing copied chat command will set actual value of IsPublic property later.
-        chat = chat with { IsPublic = false };
+        var newChatMediaId = !chat.MediaId.IsNone ? new MediaId(newChatId, chat.MediaId.LocalId) : MediaId.None;
+        var copyMedia = !newChatMediaId.IsNone && (dbChat == null || !OrdinalEquals(newChatMediaId.Value, dbChat.MediaId));
+        if (copyMedia)
+            await Commander.Call(new MediaBackend_CopyChat(newChatId, correlationId, [chat.MediaId]), true, cancellationToken)
+                .ConfigureAwait(false);
+        chat = chat with {
+            IsPublic = false,
+            MediaId = newChatMediaId
+        };
         Chat newChat;
         if (dbChat == null) {
             newChat = chat with { Id = newChatId };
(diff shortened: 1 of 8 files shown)


alexis-kochetov (approved):

src/dotnet/Chat.Service/ChatsBackend.CopyChat.cs
```diff
@@ -168,7 +168,15 @@ private async Task<Chat> CreateOrUpdateChat(string correlationId, ChatDbContext
...
         // Publishing copied chat command will set actual value of IsPublic property later.
-        chat = chat with { IsPublic = false };
+        var newChatMediaId = !chat.MediaId.IsNone ? new MediaId(newChatId, chat.MediaId.LocalId) : MediaId.None;
+        var copyMedia = !newChatMediaId.IsNone && (dbChat == null || !OrdinalEquals(newChatMediaId.Value, dbChat.MediaId));
+        if (copyMedia)
+            await Commander.Call(new MediaBackend_CopyChat(newChatId, correlationId, [chat.MediaId]), true, cancellationToken)
+                .ConfigureAwait(false);
+        chat = chat with {
+            IsPublic = false,
+            MediaId = newChatMediaId
```
frolyo, line 178: not clear how newChatMediaId is used by copied media...

src/dotnet/Core.Server/Blobs/ContentSaver.cs
```diff
@@ -19,4 +19,13 @@ public async Task Remove(string contentId, CancellationToken cancellationToken)
         await _blobStorage.Delete(contentId, cancellationToken).ConfigureAwait(false);
         await Task.Delay(PostOperationDelay, CancellationToken.None).ConfigureAwait(false);
     }
+
+    public async Task<bool> Exists(string contentId, CancellationToken cancellationToken)
+        => await _blobStorage.Exists(contentId, cancellationToken).ConfigureAwait(false);
```
frolyo, line 24: omit await

frolyo (approved):

src/dotnet/Core.Server/Blobs/ContentSaver.cs
crui3er: Why? It's reduce code amount, but it makes stack trace more readable in case of an exception. Do we have any convetion on this case?

src/dotnet/Chat.Service/ChatsBackend.CopyChat.cs
crui3er: In fact, it's the same that will be created by MediaBackend_CopyChat command.

src/dotnet/Core.Server/Blobs/ContentSaver.cs
alexis-kochetov: Usually we try to return the task directly without awaiting.

PR #1983, opened 2024-04-24
chore: fix opensearch healthcheck

AlexUstinov:


alexis-kochetov (approved):

src/dotnet/Search.Service/Module/SearchServiceModule.cs
```diff
@@ -60,7 +60,7 @@ protected override void InjectServices(IServiceCollection services)
             new SingleNodeConnectionPool(new Uri(openSearchClusterUri)),
             sourceSerializer: (builtin, settings) => new OpenSearchJsonSerializer(builtin, settings));
         services.AddSingleton<IOpenSearchClient>(_ => new OpenSearchClient(connectionSettings));
-        services.AddSingleton<OpenSearchConfigurator>()
+        services.AddSingleton(s => ActivatorUtilities.CreateInstance<OpenSearchConfigurator>(s, HostInfo.IsDevelopmentInstance))
```
frolyo: Can we use HostInfo inside OpenSearchConfigurator?

src/dotnet/Search.Service/OpenSearchConfigurator.cs
```diff
@@ -5,7 +5,7 @@
 using LogLevel = Microsoft.Extensions.Logging.LogLevel;
 namespace ActualChat.Search;
 
-public sealed class OpenSearchConfigurator(IServiceProvider services) : WorkerBase
+public sealed class OpenSearchConfigurator(IServiceProvider services, bool isDevelopmentInstance) : WorkerBase
```
frolyo: You already have HostInfo in services, let's use it.

frolyo: let's inject dependencies right from ctor

src/dotnet/Search.Service/OpenSearchConfigurator.cs
```diff
@@ -40,27 +42,26 @@ protected override async Task OnRun(CancellationToken cancellationToken)
...
-            AsyncChain.From(EnsureEntryIndexTemplate),
-            AsyncChain.From(EnsureContactIndices),
-        };
-
-        return (from chain in baseChains
-            select chain
-                .Retry(RetryDelaySeq.Exp(0, 60), 10)
-                .Log(LogLevel.Debug, Log))
-            .RunIsolated(cancellationToken);
+        await RunChain(AsyncChain.From(EnsureTemplates), cancellationToken).ConfigureAwait(false);
```
frolyo, line 48: there were also kind of Append or something like that

frolyo (approved):

src/dotnet/MLSearch.Service/Module/MLSearchServiceModule.cs
```diff
@@ -63,9 +63,10 @@ protected override void InjectServices(IServiceCollection services)
         var connectionSettings = new ConnectionSettings(
             new SingleNodeConnectionPool(new Uri(openSearchClusterUri)),
             sourceSerializer: (builtin, settings) => new OpenSearchJsonSerializer(builtin, settings));
+        services.AddSingleton<IndexNames>();
 
         services.AddSingleton<IOpenSearchClient>(_ => new OpenSearchClient(connectionSettings));
-        services.AddSingleton(e => ActivatorUtilities.CreateInstance<ClusterSetup>(e, modelGroupName))
+        services.AddSingleton(e => ActivatorUtilities.CreateInstance<ClusterSetup>(e, HostInfo.IsDevelopmentInstance, modelGroupName))
```
frolyo: both "dynamic arguments" are actually accessible from DI. Can we simplify it? I guess you want to minimize dependency scope but it's not worth it, let's stick to kind of balanced mode)

frolyo (changes requested):

src/dotnet/MLSearch.Service/Module/MLSearchServiceModule.cs
AlexUstinov: I don't want a HostInfo dependency because neither OpenSearchConfigurator nor ClusterSetup is supposed to be hosted, but modules are. However, you are right; sending just one property of HostInfo as a configuration parameter just masks the problem. Instead, configurators should receive an exact number_of_replicas setting.

frolyo (approved):

src/dotnet/Search.Service/Module/SearchServiceModule.cs
AlexUstinov: done

src/dotnet/Search.Service/OpenSearchConfigurator.cs
AlexUstinov: done

AlexUstinov: postponed

src/dotnet/Search.Service/OpenSearchConfigurator.cs
AlexUstinov: I'll keep it on mind.

src/dotnet/MLSearch.Service/Module/MLSearchServiceModule.cs
AlexUstinov: `ClusterSetup` now accepts `OpenSearchSettings` and it has `NumberOfReplicas` value.

PR #2075, opened 2024-05-14
Feat/2074 remove from place test

### LINKED ISSUE #2074: Test UserShouldNotBeListedInThePlaceChatAfterRemovingFromPlace does not pass
https://github.com/Actual-Chat/actual-chat/runs/24888784022#r0
https://github.com/Actual-Chat/actual-chat/actions/runs/9059234390/job/24888467051
Test failed because when a member is removed from place, some contacts might be missing due to event with creating author is not processed yet.

crui3er:
Close #2074


crui3er:
Destination branch is wrong

PR #2082, opened 2024-05-14
reduce recording glitches with native sample rate

alexis-kochetov:


src/dotnet/Streaming.UI.Blazor/Components/AudioRecorder/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const SAMPLES_PER_MS = 16;
+export const SAMPLES_PER_WINDOW_32 = SAMPLES_PER_MS * 32;
+export const SAMPLES_PER_WINDOW_30 = SAMPLES_PER_MS * 30;
+export const SAMPLE_RATE = 16000;
```
crui3er, line 4: Why did sample rate become lower? It was 48000 Hz, wasn't it?

src/dotnet/Streaming.UI.Blazor/Components/AudioRecorder/workers/audio-vad-worker.ts
```diff
@@ -147,19 +131,15 @@ async function processQueue(): Promise<void> {
...
-                const monoPcm = new Float32Array(resampled.buffer, 0, 512);
+                const monoPcm = new Float32Array(buffer, 0, SAMPLES_PER_WINDOW_32);
                 vadEvent = await nnVoiceDetector!.appendChunk(monoPcm);
             }
             else if (buffer.byteLength == SAMPLES_PER_WINDOW_30 * 4) {
-                // 30 ms at 48000 Hz
-                const monoPcm = new Float32Array(buffer, 0, 1440);
+                const monoPcm = new Float32Array(buffer, 0, SAMPLES_PER_WINDOW_30);
                 vadEvent = await webrtcVoiceDetector.appendChunk(monoPcm);
             }
```
crui3er, line 144: What will happen if none of above conditions are met?

crui3er (approved):

src/dotnet/Streaming.UI.Blazor/Components/AudioRecorder/constants.ts
alexis-kochetov: Microphone recording sample rate is 16K, but playback is 48K for Android and iOS by default, so when there is only one AudioContext mnicrophone stream has to be resampled, Now we use two different AudioContexts

src/dotnet/Streaming.UI.Blazor/Components/AudioRecorder/workers/audio-vad-worker.ts
alexis-kochetov: we will have to skip VAD processing for wrong sample buffers

PR #2180, opened 2024-06-13
Feat/db place

### LINKED ISSUE #2179: It should be not possible to change welcome chat type from place to private

crui3er:
Persist place data to places table
Fixes #2179 It should be not possible to change welcome chat type from place to private


src/dotnet/Chat.Service/Places.cs
```diff
@@ -95,19 +107,50 @@ public virtual async Task<ApiArray<AuthorId>> ListOwnerIds(Session session, Plac
...
+            chatDiff = !placeChange.Remove ? Change.Update(ToChatDiff(placeChange.Update.Value)) : Change.Remove<ChatDiff>();
+        }
+
+        place = await Commander.Call(changePlaceCommand, cancellationToken).ConfigureAwait(false);
+        rootChatId = rootChatId.Or(place.Id.ToRootChatId());
+
+        // Since place root chat is still used for getting place rules and place members we should synchronously update it.
+        // Should we use backend command here?
+        var chatChangeCommand = new Chats_Change(session, rootChatId, chatExpectedVersion, chatDiff);
+        placeRootChat = await Commander.Call(chatChangeCommand, true, cancellationToken).ConfigureAwait(false);
```
alexis-kochetov: Why do you use outermost command there?

alexis-kochetov: Usually we prefer backend command calls

src/dotnet/Chat.Service/Places.cs
```diff
@@ -205,6 +226,25 @@ private static ChatDiff ToChatDiff(PlaceDiff placeDiff)
...
         };
 
+    private async Task RemovePlaceChats(Session session, PlaceId placeId, CancellationToken cancellationToken)
+    {
+        var contacts = await Contacts.ListIds(session, placeId, cancellationToken).ConfigureAwait(false);
+        foreach (var contact in contacts) {
+            var chatId = contact.ChatId;
+            if (chatId.IsPeerChat(out _))
+                continue; // Ignore peer chats
+            var chat = await Chats.Get(session, chatId, cancellationToken).ConfigureAwait(false);
```
alexis-kochetov: Please add `newline` after `continue`

src/dotnet/Chat.Service/Places.cs
alexis-kochetov: Probably we should move this call to the Places backend so it can incapsulate all place-chat sync logic

alexis-kochetov (changes requested):

src/dotnet/Chat.Service/Places.cs
```diff
@@ -153,30 +196,8 @@ public virtual async Task OnPromoteToOwner(Places_PromoteToOwner command, Cancel
...
-                });
-                await Commander.Call(resetChatTagCommand, true, cancellationToken).ConfigureAwait(false);
-            }
-            var deleteChatCommand = new Chats_Change(session, chatId, null, new Change<ChatDiff> { Remove = true });
-            await Commander.Call(deleteChatCommand, true, cancellationToken).ConfigureAwait(false);
-        }
-
-        var deleteRootChatCommand = new Chats_Change(session, place.Id.ToRootChatId(), null, new Change<ChatDiff> { Remove = true });
-        await Commander.Call(deleteRootChatCommand, true, cancellationToken).ConfigureAwait(false);
+        var placeChangeCommand = new Places_Change(command.Session, command.PlaceId, null, Change.Remove<PlaceDiff>());
```
frolyo, line 176: if Places_Delete command for backward compatibility only, then maybe we need to mark it properly

frolyo (approved):

src/dotnet/Chat.Service/Places.cs
crui3er: Reworked

src/dotnet/Chat.Service/Places.cs
crui3er: Done

alexis-kochetov (approved):

PR #2196, opened 2024-06-20
Feat/create searchbot chat

andrew-kurochkin:
feat: minimal viable information on the found text
fix: bot conversation trigger should not react on system messages
fix: chats with the bot should not be indexed for the future searches


src/dotnet/MLSearch.Service/Indexing/ChatContent/ChatContentIndexWorker.cs
```diff
@@ -19,6 +20,10 @@ public async Task ExecuteAsync(MLSearch_TriggerChatIndexing job, CancellationTok
         var eventCount = 0;
         var chatId = job.Id;
 
+        bool shouldBeIndexed = await chatIndexFilter.ChatShouldBeIndexed(chatId, cancellationToken).ConfigureAwait(false);
+        if (!shouldBeIndexed)
```
AlexUstinov, line 24: We should either keep this check on job execution, or instead have it before sending 'MLSearch_TriggerChatIndexing' command to the queue.

src/dotnet/MLSearch.Service/Indexing/ChatIndexTrigger.cs
```diff
@@ -26,12 +26,11 @@ public virtual async Task OnCancelCommand(MLSearch_CancelChatIndexing e, Cancell
...
-        // TODO: The question is should we index search chats as well? (Probably yes)
-        var peerChatId = eventCommand.Entry.ChatId.PeerChatId;
-        if (peerChatId.IsNone || !peerChatId.HasUser(Constants.User.MLSearchBot.UserId)) {
-            // A normal conversation.
-            var e = new MLSearch_TriggerChatIndexing(eventCommand.Entry.ChatId);
-            await commander.Call(e, cancellationToken).ConfigureAwait(false);
-        }
+        bool shouldBeIndexed = await chatIndexFilter.ChatShouldBeIndexed(eventCommand.Entry.ChatId, cancellationToken).ConfigureAwait(false);
+        if (!shouldBeIndexed)
+            return;
```
AlexUstinov, line 31: Here we have the check before sending command, and we have another one on its execution.

src/dotnet/MLSearch.Service/Indexing/ChatContent/ChatContentIndexWorker.cs
andrew-kurochkin: This will be removed as we discussed. The current solution with filters is not future proof.

src/dotnet/MLSearch.Service/Indexing/ChatIndexTrigger.cs
andrew-kurochkin: Same as above

AlexUstinov (approved):

PR #2197, opened 2024-06-20
fix: update place background in left panel when it has changed, update place info visibility

iqmulator:


alexis-kochetov (approved):

src/dotnet/Chat.UI.Blazor/Components/LeftPanel/LeftPanelContentHeader.razor
frolyo, line 1: Looks like RenderPlace becomes larger. Imho it can be extracted to a separate component

frolyo (approved):

PR #2216, opened 2024-06-27
fix: copy trigger url, enable analytics style, add date visor

iqmulator:


src/dotnet/Chat.UI.Blazor/Components/DateVisor/date-visor.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    public dispose() {
+        if (this.disposed$.isStopped)
+            return;
+
+        this.disposed$.next();
+        this.disposed$.complete();
+    }
+
+    private onScrollHandler = () => {
```
alexis-kochetov, line 34: We have `throttle` and `debounce` methods that we use at the `virtual-list` for similar purposes - torn on somthing during scroll and turn off after. Those methods provide much more efficient way to handle timers

alexis-kochetov (changes requested):

src/dotnet/Chat.UI.Blazor/Components/DateVisor/date-visor.ts
iqmulator: Done

alexis-kochetov (approved):

PR #2239, opened 2024-07-05
feat: chat list uses virtual list

alexis-kochetov:


src/dotnet/Chat.UI.Blazor/Services/ChatListUI.cs
```diff
@@ -82,32 +78,38 @@ public int GetCountWhenLoading(ChatListKind listKind)
...
-    public virtual Task<Item> GetItem(ChatListKind listKind, int index)
+    public virtual async Task<int> IndexOf(ChatId chatId, CancellationToken cancellationToken)
     {
-        var items = GetItems(listKind);
-        lock (items) {
-            var indexIsValid = index >= 0 && index < items.Count;
-            var chatId = indexIsValid ? items[index] : ChatId.None;
-            return Task.FromResult(new Item(indexIsValid, chatId));
+        var listView = await ActiveChatListView.Use(cancellationToken).ConfigureAwait(false);
+        if (listView == null)
```
crui3er: listView has ref to PlaceId. We can use it to quickly determine if given chatId index = -1 when it is from different place.

src/dotnet/Chat.UI.Blazor/Services/ChatListUI.cs
```diff
@@ -82,32 +78,38 @@ public int GetCountWhenLoading(ChatListKind listKind)
...
 
     [ComputeMethod]
-    public virtual Task<int> GetCount(ChatListKind listKind)
+    public virtual async Task<int> GetCount(PlaceId placeId, CancellationToken cancellationToken)
     {
-        var items = GetItems(listKind);
-        lock (items)
-            return Task.FromResult(items.Count);
+        var listView = await ActiveChatListView.Use(cancellationToken).ConfigureAwait(false);
+        if (listView == null)
```
crui3er: listView has ref to PlaceId. We can use it to quickly determine if count = 0 when given placeId does not correspone active list view placeId.

crui3er (approved):

PR #2262, opened 2024-07-16
Index and filter over chat metadata

AlexUstinov:


src/dotnet/MLSearch.Service/Engine/ChatFilter.cs
```diff
@@ -0,0 +1,16 @@
+namespace ActualChat.MLSearch.Engine;
+
+internal enum InclusionMode
```
andrew-kurochkin, line 6: I do not understand IncludeStrictly vs Include concept. Feels very unnatural and will be a source of future bugs. I would think that we have to think through this.

andrew-kurochkin: I believe you can hide this abstraction inside the ChatFilter itself and make it private so it doesn't leak anywhere. In this case ChatFilter can have factory methods that would incapsulate concepts you wanted. Something like ChatFilter.OnlyPublic() vs new ChatFilter.

src/dotnet/MLSearch.Service/Engine/IQueryBuilder.cs
```diff
@@ -8,4 +10,9 @@ internal interface IQueryBuilder
     void ApplyRangeFilter(Int32RangeFilter rangeFilter);
     void ApplyRangeFilter(Int64RangeFilter rangeFilter);
     void ApplyRangeFilter(DateRangeFilter rangeFilter);
+    void ApplyKeywordFilter<TDocument>(KeywordFilter<TDocument> keywordFilter)
```
andrew-kurochkin, line 13: It is super confusing to have KeywordFilter and FreeText. I do believe we have to be more explicit here. On top of that I believe the scoring weight should be a passed parameter along with the filter. Also, if we do these two changes I don't see the difference if we use Query Builder directly vs having another abstraction that effectively doesn't hide any complexity.
I don't like an idea of adding abstractions when not needed. What you think?

src/dotnet/MLSearch.Service/Engine/OpenSearch/Setup/ClusterSetupActions.cs
```diff
@@ -26,7 +26,8 @@ internal sealed class ClusterSetupActions(
     Tracer baseTracer
 ) : IClusterSetupActions
 {
-     private readonly Tracer _tracer = baseTracer[typeof(ClusterSetup)];
+    private const string TimestampFieldName = "timestamp";
```
andrew-kurochkin, line 29: It is a timestamp of an entry when it was indexed, isn't it? 
I believe we should change the name to be more explicit regarding the event when this timestamp is taken.

andrew-kurochkin (changes requested):

andrew-kurochkin (approved):
Talked offline regarding changes needed. Decided to postpone some changes for the later pr.

PR #2285, opened 2024-07-24
feat: monitor firebase messaging token changes and update at the server

alexis-kochetov:


src/dotnet/App.Maui/Platforms/Android/Notifications/FirebaseMessagingService.cs
```diff
@@ -38,6 +44,17 @@ public class FirebaseMessagingService : Firebase.Messaging.FirebaseMessagingServ
     public override void OnNewToken(string token)
     {
         Log.LogDebug("OnNewToken: '{Token}'", token);
+        if (TryGetScopedServices(out var services)) {
```
crui3er: ScopedServices are initialized when MauiBlazorApp is initialized, so UI is loaded. For case when we have UI loaded, NotificationsPermissionBanner invokes registering device.
If we want to react on new token message when ui is not loaded we need to use services from root container only.
Also I think we should check if notification permission is granted.

crui3er (changes requested):

src/dotnet/App.Maui/Platforms/Android/Notifications/FirebaseMessagingService.cs
alexis-kochetov: We don't have a session in the root container, so do you have any ideas 🤔

src/dotnet/App.Maui/Platforms/iOS/AppDelegate.cs
```diff
@@ -42,6 +52,28 @@ public override void DidEnterBackground(UIApplication application)
...
+            return;
+
+        var rpcHub = appServices.GetService<RpcHub>();
+        var commander = appServices.GetService<ICommander>();
+        var sessionResolver = appServices.GetRequiredService<TrueSessionResolver>();
+        if (rpcHub != null && commander != null)
+            _ = Task.Run(async () => {
+                await rpcHub.WhenClientPeerConnected().ConfigureAwait(false);
+                var session = await sessionResolver.GetSession().ConfigureAwait(false);
+                var command = new Notifications_RegisterDevice(session, token, DeviceType.AndroidApp);
```
crui3er: Use DeviceType.iOSApp

src/dotnet/App.Maui/Platforms/iOS/AppDelegate.cs
```diff
@@ -42,6 +52,28 @@ public override void DidEnterBackground(UIApplication application)
...
+    public void DidReceiveRegistrationToken (Firebase.CloudMessaging.Messaging messaging, string fcmToken)
+    {
+        // Monitor token generation: To be notified whenever the token is updated.
+        var token = fcmToken;
+        Log.LogDebug("OnNewToken: '{Token}'", token);
+        var appServices = IPlatformApplication.Current?.Services;
+        if (appServices == null)
+            return;
+
+        var rpcHub = appServices.GetService<RpcHub>();
```
crui3er: I think we can extract code block below and use both from Android and iOS platforms.

src/dotnet/App.Maui/Platforms/iOS/AppDelegate.cs
```diff
@@ -42,6 +52,28 @@ public override void DidEnterBackground(UIApplication application)
...
+        var commander = appServices.GetService<ICommander>();
+        var sessionResolver = appServices.GetRequiredService<TrueSessionResolver>();
+        if (rpcHub != null && commander != null)
+            _ = Task.Run(async () => {
+                await rpcHub.WhenClientPeerConnected().ConfigureAwait(false);
```
crui3er: Add exception handling for background task and log an error if it occurs. May be, make sense to use our BackgrounTask.

crui3er (changes requested):

PR #2329, opened 2024-08-06
Bugfix: notification device should be registered with auth user

alexis-kochetov:
…only

### DIFF
--- a/src/dotnet/Api.Contracts/Notification/INotifications.cs
+++ b/src/dotnet/Api.Contracts/Notification/INotifications.cs
@@ -15,6 +15,8 @@ public interface INotifications : IComputeService
     [CommandHandler]
     Task OnRegisterDevice(Notifications_RegisterDevice command, CancellationToken cancellationToken);
     [CommandHandler]
+    Task OnDeregisterDevice(Notifications_DeregisterDevice command, CancellationToken cancellationToken);
+    [CommandHandler]
     Task OnNotifyMembers(Notifications_NotifyMembers command, CancellationToken cancellationToken);
 }
 
@@ -32,6 +34,14 @@ public sealed partial record Notifications_RegisterDevice(
     [property: DataMember, MemoryPackOrder(1)] Symbol DeviceId,
     [property: DataMember, MemoryPackOrder(2)] DeviceType DeviceType
 ) : ISessionCommand<Unit>, IApiCommand;
+
+[DataContract, MemoryPackable(GenerateType.VersionTolerant)]
+// ReSharper disable once InconsistentNaming
+public sealed partial record Notifications_DeregisterDevice(
+    [property: DataMember, MemoryPackOrder(0)] Session Session,
+    [property: DataMember, MemoryPackOrder(1)] Symbol DeviceId
+) : ISessionCommand<Unit>, IApiCommand;
+
 
 [DataContract, MemoryPackable(GenerateType.VersionTolerant)]
 // ReSharper disable once InconsistentNaming
--- a/src/dotnet/App.Maui/Platforms/Android/AndroidDeviceTokenRetriever.cs
+++ b/src/dotnet/App.Maui/Platforms/Android/AndroidDeviceTokenRetriever.cs
@@ -12,7 +12,7 @@ public class AndroidDeviceTokenRetriever(IServiceProvider services) : IDeviceTok
     public async Task<string?> GetDeviceToken(CancellationToken cancellationToken)
     {
         try {
-            var javaString = await FirebaseMessaging.Instance.GetToken().AsAsync<Java.Lang.String>().ConfigureAwait(true);
+            var javaString = await FirebaseMessaging.Instance.GetToken().AsAsync<Java.Lang.String>().ConfigureAwait(false);
             var token = javaString.ToString();
             Log.LogDebug("FCM token is \'{Token}\'", token);
             return token;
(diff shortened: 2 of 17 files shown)


src/dotnet/Api.Contracts/Notification/INotifications.cs
```diff
@@ -33,6 +35,14 @@ public sealed partial record Notifications_RegisterDevice(
     [property: DataMember, MemoryPackOrder(2)] DeviceType DeviceType
 ) : ISessionCommand<Unit>, IApiCommand;
 
+[DataContract, MemoryPackable(GenerateType.VersionTolerant)]
+// ReSharper disable once InconsistentNaming
+public sealed partial record Notifications_DeregisterDevice(
```
crui3er, line 40: May be, better use 'Unregister. It sounds more familiar. Just like an idea.

src/dotnet/App.Maui/Platforms/Android/AndroidDeviceTokenRetriever.cs
```diff
@@ -22,4 +22,14 @@ public class AndroidDeviceTokenRetriever(IServiceProvider services) : IDeviceTok
             return null;
         }
     }
+
+    public async Task DeleteDeviceToken(CancellationToken cancellationToken)
+    {
+        try {
+            await FirebaseMessaging.Instance.DeleteToken().AsAsync().ConfigureAwait(true);
```
crui3er: Do we need to continueOnCapturedContext?

src/dotnet/Notification.Service/Notifications.cs
```diff
@@ -57,6 +58,21 @@ public class Notifications(IServiceProvider services) : INotifications
         await Commander.Run(registerDeviceCommand, cancellationToken).ConfigureAwait(false);
     }
 
+    // [CommandHandler]
+    public virtual async Task OnDeregisterDevice(
+        Notifications_DeregisterDevice command, CancellationToken cancellationToken)
+    {
+        var (session, deviceId) = command;
+        var account = await Accounts.GetOwn(session, cancellationToken).ConfigureAwait(false);
+        var existingDevices = await Backend.ListDevices(account.Id, cancellationToken).ConfigureAwait(false);
```
crui3er, line 67: I think that deviceId is primary key for device and we should get device by id and then remove it if it exists.
Also we can log warn if existent device userId is different from the account id.

src/dotnet/Notification.UI.Blazor/Module/NotificationBlazorUIModule.cs
```diff
@@ -15,6 +16,7 @@ protected override void InjectServices(IServiceCollection services)
     {
         services.AddFusion();
         services.AddScoped<NotificationUI>();
+        services.AddScoped<INotificationUI, NotificationUI>(c => c.GetRequiredService<NotificationUI>());
```
crui3er: We have more short syntax for the same:
services.AddAlias<INotificationUI, NotificationUI>(ServiceLifetime.Scoped)

src/dotnet/Notification.UI.Blazor/NotificationUI.cs
```diff
@@ -132,29 +132,70 @@ public void SetPermissionState(string permissionState)
...
+    public async Task EnsureDeviceRegistered(CancellationToken cancellationToken = default)
+    {
+        var deviceId = await DeviceTokenRetriever.GetDeviceToken(cancellationToken).ConfigureAwait(false);
+        var existingTask = _registerDeviceTask;
+        if (existingTask != null) {
+            var alreadyRegisteredDeviceId = await existingTask.ConfigureAwait(false);
+            if (alreadyRegisteredDeviceId == deviceId)
+                return;
+        }
+        _registerDeviceTask = null;
```
crui3er: you need to synchronize write access to _registerDeviceTask field

src/dotnet/UI.Blazor/Services/AccountUI/AccountUI.cs
```diff
@@ -61,6 +63,20 @@ public TimeSpan GetPostChangeInvalidationDelay(TimeSpan maxInvalidationDelay)
         return (changedAt + maxInvalidationDelay - CpuClock.Now).Positive();
     }
 
-    public Task SignOut()
-        => ClientAuth.SignOut();
+    public async Task SignIn(string schema)
+    {
+        await ClientAuth.SignIn(schema).ConfigureAwait(false);
```
crui3er, line 68: ClientAuth.SignIn is used on ProviderSelectStep. Don't we need to replace it with AccountUI.SignIn?

crui3er (changes requested):

src/dotnet/Api.Contracts/Notification/INotifications.cs
alexis-kochetov: Unregister has different meaning - https://english.stackexchange.com/questions/25931/unregister-vs-deregister

src/dotnet/App.Maui/Platforms/Android/AndroidDeviceTokenRetriever.cs
alexis-kochetov: AI code completion is a mess!

src/dotnet/Notification.Service/Notifications.cs
alexis-kochetov: I think we should be able to deregister only own device Ids - that's why I'm checking the list of existing devices for the user.

src/dotnet/Notification.UI.Blazor/NotificationUI.cs
alexis-kochetov: OK. it will be simpler

PR #2331, opened 2024-08-06
Hide TimeoutException from being intercepted by vscode's 'All Exceptions' breakpoint

AlexUstinov:


src/dotnet/Core.Server/Mesh/MeshWatcher.cs
```diff
@@ -96,16 +96,22 @@ protected override async Task OnRun(CancellationToken cancellationToken)
...
-                    cancellationToken.ThrowIfCancellationRequested();
-                    if (!canRead)
-                        throw new OperationCanceledException("Subscription to changes is lost.");
-                    consumeTask = null;
+                    if (canReadResult.HasError) {
+                        if (canReadResult.Error is not TimeoutException)
+                            canReadResult.ThrowIfError();
+                    }
+                    else {
+                        var canRead = canReadResult.Value;
```
alexyakunin: It's better to use `result.IsValue(out var value, out var error)` pattern here.

alexyakunin (changes requested):

src/dotnet/Core.Server/Mesh/MeshWatcher.cs
AlexUstinov: Done. But previous version has already sneaked into the dev branch through another PR.

alexyakunin (approved):

PR #2343, opened 2024-08-07
feat: separate place index #2336

### LINKED ISSUE #2336: Split chat and place indices

frolyo:


src/dotnet/Chat.Contracts/ChatsBackendExt.cs
```diff
@@ -140,15 +140,49 @@ public static class ChatsBackendExt
...
+
+            yield return chats;
+
+            var last = chats[^1];
+            lastId = last.Id;
+            minVersion = last.Version;
+        }
+    }
+
+    public static async IAsyncEnumerable<ApiArray<Chat>> BatchChangedGroups(
```
crui3er, line 141: What do you need both BatchChanged and BatchChangedGroups methods? They looks identical.

src/dotnet/Chat.Service/Db/DbChat.cs
```diff
@@ -7,7 +7,7 @@ namespace ActualChat.Chat.Db;
 
 [Table("Chats")]
 [Index(nameof(CreatedAt))]
-[Index(nameof(Version), nameof(Id))]
+[Index(nameof(Version), nameof(Id), nameof(IsPlaceRootChat))]
```
crui3er: Can't strictly prove it, but I have an impression that we need only index by Version field with included Id and IsPlaceRootChat. I think inculding Id as second prop in the index does not much improve ordering by Id.
Indexes with included fields can be defined the following way https://learn.microsoft.com/en-us/ef/core/modeling/indexes?tabs=data-annotations#included-columns
It's discussable which kind of index we should use to improve ListChanged query performance. So you keep it as it is. It's just for imformation.

crui3er (changes requested):

src/dotnet/Chat.Contracts/ChatsBackendExt.cs
frolyo: Removed the first one since it's not even used

src/dotnet/Chat.Service/Db/DbChat.cs
frolyo: done, please check

crui3er (approved):

PR #2355, opened 2024-08-12
Feat/chat bot context

andrew-kurochkin:


src/dotnet/MLSearch.Service/Bot/External/ExternalChatBotConversationHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+        if (lastUpdatedDocument.AuthorId == botId)
+            return;
+        if (lastUpdatedDocument.Kind != ChatEntryKind.Text) {
+            // Can't react on anything besides text yet.
+            return;
+        }
+        
+        /// Minimal WebHook implementation.
+        
+        HttpClient client = new HttpClient();
```
AlexUstinov: It is recommended to instantiate one instance per application, not one instance per use.

src/dotnet/MLSearch.Service/Bot/External/ExternalChatBotConversationHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+        
+        // Note: 
+        // The outmost "input" is part of the /invoke method of the langserve.
+        // There are also "kwargs" and "config" parts available.
+        // The inner <input> structure must be syncronized with the prompt used on the bot side.
+        requestMessage.Content = JsonContent.Create(new {
+            input = lastUpdatedDocument.Content
+        });
+        botToolsContextHandler.SetContext(requestMessage, conversationId: chatId);
+        HttpResponseMessage response = client.SendAsync(requestMessage, cancellationToken).GetAwaiter().GetResult();
```
AlexUstinov: Why do you wait synchronously?

src/dotnet/MLSearch.Service/Bot/Tools/BotAuthSchemeHandler.cs
```diff
@@ -0,0 +1,79 @@
...
+    }
+    public ClaimsPrincipal GetValidatedClaims(HttpRequest request) {
+        const string BEARER_PREFIX = "Bearer ";
+        // TODO: Read the token from request headers/cookies
+        // Check that it's a valid session, depending on your implementation
+        string bearerToken = request.Headers.Authorization.Where(e=> e.StartsWith(BEARER_PREFIX)).FirstOrDefault();
+        if (bearerToken == null) {
+            return null;
+        }
+        var claims = VerifyToken(bearerToken.Substring(BEARER_PREFIX.Length));
```
AlexUstinov: Here must be a warning about proper string comparison. I think StringComparison.Ordinal is what being expected here (and in the StartsWith call above).

src/dotnet/MLSearch.Service/Bot/Tools/BotAuthSchemeHandler.cs
```diff
@@ -0,0 +1,79 @@
...
+            //IssuerSigningKey = this.Options.SigningCredentials.Key,
+            IssuerSigningKey = this.StubSigningCredentials.Key,
+            ClockSkew = TimeSpan.Zero
+        };
+
+        var tokenHandler = new JwtSecurityTokenHandler();
+        try
+        {
+            SecurityToken validatedToken = null;
+            var claims = tokenHandler.ValidateToken(token, validationParameters, out validatedToken);
```
AlexUstinov: Why not to use `out var` instead of explicit declaration?

src/dotnet/MLSearch.Service/Bot/Tools/Context/BotToolsContextHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+public class BotToolsContextHandler(IOptionsMonitor<BotToolsContextHandlerOptions> options): IBotToolsContextHandler
+{
+    private const string BEARER_PREFIX = "Bearer ";
+
+    public IBotToolsContext GetContext(HttpRequest request) => new BotToolsContext(GetValidatedClaims(request));
+    public void SetContext(HttpRequestMessage request, string conversationId) {
+        var config = options.CurrentValue;
+        var claims = new List<Claim>();
+        claims.Add(new Claim(BotToolsContext.ConversationClaimType, conversationId));
+        var expiration = DateTime.Now.Add(config.ContextLifetime);
```
AlexUstinov: It is better not to use DateTime.Now but instead use some clock (e.g. MomentClock) provided as constructor parameter.
Also, what about timezone? Now takes time in the current server timezone. If token receiver uses other timezone we may have some unexpected behavior when token lives longer than expected or expires sooner.

src/dotnet/MLSearch.Service/Bot/Tools/Context/BotToolsContextHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+            signingCredentials: config.SigningCredentials
+        );
+
+        var signedToken = new JwtSecurityTokenHandler().WriteToken(token);
+        request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", signedToken);
+    }
+
+    private ClaimsPrincipal? GetValidatedClaims(HttpRequest request) {
+        string? bearerToken = request.Headers.Authorization
+            .Where(e => e?.StartsWith(BEARER_PREFIX, StringComparison.InvariantCulture) == true)
```
AlexUstinov: StringComparison.Ordinal is faster than InvariantCulture

src/dotnet/MLSearch.Service/Bot/External/ExternalChatBotConversationHandler.cs
andrew-kurochkin: Have checked notes regarding the HttpClient class. Agree.

src/dotnet/MLSearch.Service/Bot/External/ExternalChatBotConversationHandler.cs
andrew-kurochkin: An artifact from testing. Will fix.

src/dotnet/MLSearch.Service/Bot/Tools/BotAuthSchemeHandler.cs
andrew-kurochkin: will fix.

src/dotnet/MLSearch.Service/Bot/Tools/BotAuthSchemeHandler.cs
andrew-kurochkin: hm. new syntax for me. will use.

src/dotnet/MLSearch.Service/Bot/Tools/Context/BotToolsContextHandler.cs
andrew-kurochkin: - The DateTime expiration argument is always converted into NumericDate format in the JwtSecurityToken internally. (seconds since epoch).
- Since it is seconds there's no need to use MomentClock or any other precise clocks.
- It can't be a parameter in the constructor as it should use current time.

src/dotnet/MLSearch.Service/Bot/Tools/Context/BotToolsContextHandler.cs
andrew-kurochkin: ok

src/dotnet/MLSearch.Service/Bot/Tools/Context/BotToolsContextHandler.cs
AlexUstinov: - As datetime doesn't store timezone information internal tick count of `DateTime.Now` value depends on server timezone. Literally it offsets UtcNow to server timezone (see https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/DateTime.cs#L1520).
So `JwtSecurityToken` has to internally offset time back to Utc (See https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L274 and https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/DateTimeUtil.cs#L85). While it works those back-and-forth conversion are avoidable should we use `DateTime.UtcNow`.
- I mentioned `MomentClock` passed via constructor parameter because it makes code testable as you are not stick to real time and can control time flow via the clock mock. This allows to get rid of flucky tests.

src/dotnet/MLSearch.Service/Engine/SearchQuery.cs
```diff
@@ -1,21 +1,21 @@
 
 namespace ActualChat.MLSearch.Engine;
 
-internal sealed class SearchQuery
+public sealed class SearchQuery
```
AlexUstinov, line 4: Why has everything turned to public? Is this avoidable?

src/dotnet/MLSearch.Service/Bot/External/ExternalChatBotConversationHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+        if (updatedDocuments == null) {
+            return;
+        }
+
+        var lastUpdatedDocument = updatedDocuments.LastOrDefault();
+        if (lastUpdatedDocument == null)
+            return;
+
+        var chatId = lastUpdatedDocument.ChatId;
+        AuthorId botId = new(chatId, Constants.User.MLSearchBot.AuthorLocalId, AssumeValid.Option);
```
AlexUstinov, line 61: I noticed some inconsistencies in code style:
1. Inconsistent declaration of variables here

src/dotnet/MLSearch.Service/Bot/External/ExternalChatBotConversationHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+        if (!currentSettings.IsEnabled) {
+            return;
+        }
+        if (updatedDocuments == null) {
+            return;
+        }
+
+        var lastUpdatedDocument = updatedDocuments.LastOrDefault();
+        if (lastUpdatedDocument == null)
+            return;
```
AlexUstinov: 2. Inconsistent `if` braces as first if has braces but second does not.

src/dotnet/MLSearch.Service/Bot/Tools/ConversationToolsController.cs
```diff
@@ -0,0 +1,93 @@
...
+[Produces("application/json")]
+public sealed class ConversationToolsController(ICommander commander, IBotToolsContextHandler botToolsContext, UrlMapper urlMapper): ControllerBase
+{
+    public sealed class Reply {
+        public required string Text {get; init;}
+    }
+
+    public sealed class ForwardLocalLinks {
+        public string? Comment {get; set;}
+        public required List<string> Links { get; init; }
```
AlexUstinov: 3. Inconsistent spaces in auto property declaration. (`{get; set;}` vs `{ get; init; }`)

src/dotnet/MLSearch.Service/Bot/Tools/Context/BotToolsContextHandler.cs
andrew-kurochkin: I agree that DateTime.UtcNow is better. I'll change into that.
I generally do not disagree with the proposal to add a source of DateTime to a constructor where it make sense. However this is not the case here. Let me try to explain my thoughts:
- This class has two sides: one generates a token, and the second one is checking its validity.
- If we pass a source of a DateTime anywhere (a constructor / a parameter) it must be used by both methods.
- It's very easy to set an expiration time for the token, however the validation counterpart is not that easy. I've double-checked it and they use DateTime.UtcNow in it's inner static method: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/eb4df8fc5a9a4fbf50de2dd5bfd263a225c7c87a/src/Microsoft.IdentityModel.Tokens/ValidatorUtilities.cs#L39

src/dotnet/MLSearch.Service/Engine/SearchQuery.cs
andrew-kurochkin: Let me state the reason first:
services.AddMvcCore().AddApplicationPart(GetType().Assembly);
Apparently this method scans for public controllers only. It silently ignores all controllers declared internal.
It is possible to declare public controller classes with internal constructors and skip primary constructors all together. Not sure what is better here. Will discuss offline

andrew-kurochkin: Apparently controllers and its main constructor must be public. This forces all arguments being public.

AlexUstinov (approved):

PR #2371, opened 2024-08-15
Feat/2370 show index doc id

### LINKED ISSUE #2370: Show index doc ids to debug ChatContentArranger

crui3er:
Closes #2370
image
image

### DIFF
--- a/src/dotnet/MLSearch.Service/MLSearch.cs
+++ b/src/dotnet/MLSearch.Service/MLSearch.cs
@@ -1,9 +1,39 @@
 ﻿using ActualChat.Chat;
+using ActualChat.Users;
 
 namespace ActualChat.MLSearch;
 
-public class MLSearchImpl (ICommander commander): IMLSearch
+public class MLSearchImpl (IServiceProvider services): IMLSearch
 {
+    private IChats? _chats;
+    private IAccounts? _accounts;
+
+    private IChats Chats => _chats ??= services.GetRequiredService<IChats>(); // Lazy resolving to prevent cyclic dependency
+    private IAccounts Accounts => _accounts ??= services.GetRequiredService<IAccounts>(); // Lazy resolving to prevent cyclic dependency
+    private IMLSearchBackend Backend { get; } = services.GetRequiredService<IMLSearchBackend>();
+
+    private ICommander Commander { get; } = services.Commander();
+
+    public virtual async Task<string> GetIndexDocIdByEntryId(
+        Session session,
+        ChatEntryId chatEntryId,
+        CancellationToken cancellationToken)
+    {
+        if (session.Id.IsEmpty)
+            throw new ArgumentOutOfRangeException(nameof(session));
+
+        if (chatEntryId.IsNone)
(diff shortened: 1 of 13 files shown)


src/dotnet/MLSearch.Service/MLSearch.cs
```diff
@@ -1,9 +1,39 @@
 ﻿using ActualChat.Chat;
+using ActualChat.Users;
 
 namespace ActualChat.MLSearch;
 
-public class MLSearchImpl (ICommander commander): IMLSearch
+public class MLSearchImpl (IServiceProvider services): IMLSearch
```
AlexUstinov: :( I tried to prevent service locator pattern in the ML module.
This makes code untestable.

src/dotnet/MLSearch.Service/MLSearchBackend.cs
```diff
@@ -0,0 +1,23 @@
+using ActualChat.MLSearch.Indexing.ChatContent;
+
+namespace ActualChat.MLSearch;
+
+public class MLSearchBackend(IServiceProvider services) : IMLSearchBackend
+{
+    internal IChatContentDocumentLoader DocumentLoader { get; } =
```
AlexUstinov: Can we avoid using service locator?

src/dotnet/MLSearch.Service/MLSearch.cs
crui3er: Yep. I understand, that explicit dependecies better than service locator. Fixed.

src/dotnet/MLSearch.Service/MLSearchBackend.cs
crui3er: Fixed

AlexUstinov (approved):

src/dotnet/MLSearch.Service/MLSearch.cs
```diff
@@ -1,9 +1,36 @@
...
 namespace ActualChat.MLSearch;
 
-public class MLSearchImpl (ICommander commander): IMLSearch
+public class MLSearchImpl (ICommander commander, IMLSearchBackend backend, IChats chats, IAccounts accounts): IMLSearch
 {
+    private IChats Chats => chats;
+    private IAccounts Accounts => accounts;
+    private IMLSearchBackend Backend { get; } = backend;
+
+    private ICommander Commander { get; } = commander;
```
AlexUstinov, line 12: It is totally fine to have such a private proxy property for primary constructor parameter. However, I personally decided not to use this pattern. The main reason behind that is both primary constructor parameter and proxy property are accessible for class members, so each time you want to use e.g. `ICommander` instance you have to select between `commander` and `Commander`. So, highly likely, over the time the code will access `ICommander` both ways bringing some degree of inconsistency.

frolyo: this property duplicates commander, can we just remove it?

frolyo (approved):

PR #2382, opened 2024-08-20
Feat/bot conversation

andrew-kurochkin:
This PR unlocks some of the bot functionality.
- Enable Thread Id for the LangGraph.
- Temporary disabling LangFuse use. This is possible since prompt is not used after the migration to the Claude v3 model.
- Temporary removing the prompt. Will be re-integrated with the LangFuse.


services/chatbot/app/server.py
```diff
@@ -95,10 +96,22 @@ def _extract_thread_id(
...
+    jwt_bearer_token = request.headers.get("Authorization", None)
+    assert(jwt_bearer_token is not None, "Authorization header must be set")
+    assert(jwt_bearer_token.startswith("Bearer "), "Sanity check")
+    jwt_token = jwt_bearer_token.replace("Bearer ", "", 1)
+
+    claims = jwt.decode(
+        jwt_token,
+        options = {
+            # TODO: Validate JWT token
+            "verify_signature": False
```
andrew-kurochkin, line 109: JWT verification is temporary disabled due to issues with the different cryptographic elliptic curve primitives in .Net and Python libraries.

AlexUstinov (approved):

PR #2397, opened 2024-08-27
fix: system messages in digest

evgeniy-ac:


src/dotnet/Users.Service/EmailsBackend.cs
```diff
@@ -122,30 +122,36 @@ private async Task<DigestParameters> FindUnreadChats(
...
             var digestChat = new DigestParameters.DigestChat {
                 Name = chat.Title,
                 Link = UrlMapper.ToAbsolute(Links.Chat(chat.Id)),
                 UnreadCount = maxEntryId - chatPosition.EntryLid,
                 FirstUnreadChatEntry = new DigestParameters.DigestChatEntry {
                     At = DeltaText.Get(userBeginsAt, userNow).Text,
                     AuthorName = author.Avatar.Name,
-                    Text = chatEntry.Content.IsNullOrEmpty()
+                    Text = firstUnreadMessage.Content.IsNullOrEmpty()
                         ? "N/A"
```
crui3er, line 153: I have a remark that does not connected directly with this PR.
"N/A" looks ugly. What we can show instead: description of attachment or may be better to look further to the next entry with meaningful content.

crui3er (approved):

src/dotnet/Users.Service/EmailsBackend.cs
evgeniy-ac: Agree. We will do it eventually

PR #2419, opened 2024-09-05
fix: active-recording-svg, update RecorderToggle, fix audio-panel css

iqmulator:


frolyo (approved):

alexis-kochetov (changes requested):
Inefficient, will have to refactor and optimize.  I will do it myself

PR #2423, opened 2024-09-06
Feat/private chat search

AlexUstinov:


src/dotnet/Chat.Service/ChatsBackend.cs
```diff
@@ -123,6 +123,52 @@ public virtual async Task<ApiArray<ChatId>> GetPublicChatIdsFor(PlaceId placeId,
         return sChatIds.Select(x => new ChatId(x)).Where(x => !x.IsPlaceRootChat).ToApiArray();
     }
 
+    // [ComputeMethod]
+    public virtual async Task<ApiArray<ChatId>> GetPrivateChatIdsForUser(UserId userId, PlaceId? placeId, CancellationToken cancellationToken)
```
frolyo: we search for user chats via ContactsBackend, not directly through ChatsBackend.

frolyo: please take a look at already existing ContactsBackend.ListIdsForSearch, ...
https://github.com/Actual-Chat/actual-chat/blob/a1b841745c7e0983fd41badb1ded7ba2808f038f/src/dotnet/Contacts.Service/ContactsBackend.cs#L105

frolyo (changes requested):

src/dotnet/MLSearch.Service/Engine/Filters.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+        else {
+            ChatFilter.IncludePublic = true;
+        }
+        return ValueTask.CompletedTask;
+    }
+
+    internal async ValueTask IncludePrivate(UserId userId, PlaceId? placeId, CancellationToken cancellationToken)
+    {
+        var privateContacts = await contacts.ListIdsForSearch(userId, placeId, cancellationToken).ConfigureAwait(false);
```
frolyo: though it will work, I guess you need only private chats. This method returns both private and public afair.

frolyo (approved):

src/dotnet/MLSearch.Service/Engine/Filters.cs
AlexUstinov: You are right, will fix.

frolyo (approved):

PR #2509, opened 2024-09-25
fix: message from peer chats were not found in place #2491

### LINKED ISSUE #2491: Search for messages in place does not consider peer chats

frolyo:


src/dotnet/Core/Constants.DebugMode.cs
```diff
@@ -58,7 +58,7 @@ public static bool AudioPlaybackPlayMyOwnAudio
         public static readonly bool TranscriberGoogle = false;
 
         // Search
-        public static readonly bool OpenSearchRequest = false;
+        public static readonly bool OpenSearchRequest = true;
```
crui3er: Do you enable debug mode for OpenSearch on purpose?

crui3er (approved):

src/dotnet/Core/Constants.DebugMode.cs
frolyo: fixed, thanks

PR #2554, opened 2024-10-03
feat: show empty search results #2527

### LINKED ISSUE #2527: Show empty search results

frolyo:


src/dotnet/UI.Blazor.App/Components/NoResults/NoResults.razor
```diff
@@ -0,0 +1,11 @@
+@namespace ActualChat.UI.Blazor.App.Components
```
crui3er, line 1: Why did you put here? Not next to FoundChatList?

src/dotnet/UI.Blazor.App/Components/NoResults/no-results-svg.lit.ts
```diff
@@ -0,0 +1,51 @@
+import { customElement } from 'lit/decorators.js';
+import { html, LitElement } from 'lit';
+
+@customElement('no-results-svg')
```
crui3er, line 4: Why is this svg wrapped into a lit component?

frolyo: the same way as the other not found views

src/dotnet/UI.Blazor.App/Components/NoResults/NoResults.razor
frolyo: It's just a special cat component, so I decided to put it along with other cat components

PR #2559, opened 2024-10-04
fix: After first sign in, user is not redirected to '/chat/' page #2558

### LINKED ISSUE #2558: After first sign in, user is not redirected to '/chat/' page
Closing SignInModal cause NavigateBack that discard redirect to '/chat/' initiated right before.

crui3er:
Closes #2558


src/dotnet/UI.Blazor/Services/AccountUI/AccountUI.cs
```diff
@@ -11,7 +11,9 @@ public partial class AccountUI : ScopedWorkerBase<UIHub>, IComputeService, INoti
     private readonly MutableState<Moment> _lastChangedAt;
     private readonly MutableState<SignInRequest?> _activeSignInRequest;
     private readonly TimeSpan _maxInvalidationDelay;
+    private readonly object _postponeOnSignedInWorkflowTasksSyncObject = new();
```
frolyo: we agreed to name it lock, instead of SyncObject

crui3er: Ok. May be, I missed it.

frolyo (approved):

PR #2560, opened 2024-10-04
feat: add summarization to digest

evgeniy-ac:


src/dotnet/Users.Service/EmailsBackend.cs
```diff
@@ -65,16 +62,14 @@ await EmailSender.Send(
...
-        TimeZoneInfo timeZoneInfo,
         CancellationToken cancellationToken)
     {
         const int takeChats = 5;
         var totalUnreadCount = 0;
         var unreadChats = new List<DigestParameters.DigestChat>();
         var accountSettings = ServerKvasBackend.GetUserClient(account.Id);
         var contactIds = await ContactsBackend
-            .ListIdsForSearch(account.Id, null, false, cancellationToken)
+            .ListIds(account.Id, PlaceId.None, cancellationToken)
```
crui3er: Why do you ignore place chats?

crui3er (approved):

src/dotnet/Users.Service/EmailsBackend.cs
evgeniy-ac: returned the old one back

alexis-kochetov (approved):

PR #2569, opened 2024-10-07
fix: notify btn for text-mode, hide tabs for place

iqmulator:


src/dotnet/UI.Blazor.App/Components/ChatMessageEditor/chat-message-editor.css
```diff
@@ -168,18 +168,16 @@ body.hoverable .post-panel-btn:hover {
...
 .default-mode .notify-call-btn {
     @apply static right-auto;
     @apply flex;
     @apply opacity-100;
-    @starting-style {
-        opacity: 0;
-    }
+}
+.text-mode .post-panel-btn.notify-call-btn {
+    @apply hidden;
```
crui3er, line 180: This is OK. I think we can also get rid of old behavior: that this button movы from right position to bottom left when editor gets focused and editor has text.

src/dotnet/UI.Blazor.App/Components/Navbar/NavbarContent.razor
```diff
@@ -12,6 +12,9 @@
     var isAdministration = NavbarUI.SelectedGroupId == NavbarGroupIds.Administration;
     var isTestPages = NavbarUI.SelectedGroupId == NavbarGroupIds.TestPages;
     var cls = isChats ? "navbar-chats" : "navbar-pages";
+    if (!placeId.IsNone) {
```
crui3er: This is not needed.

src/dotnet/UI.Blazor.App/Components/ChatList/chat-list.css
```diff
@@ -254,6 +254,9 @@ body.hoverable .chat-list .navbar-item:hover .navbar-item-ending .slot.hoverable
     @apply mx-0;
     @apply text-primary-title;
 }
+.navbar-group.place .tab-panel .tab-panel-tabs {
```
crui3er: This is not needed.
We should hide tabs only when search mode is on.
In chats it happens because we cover this area with search results.

crui3er (changes requested):

alexis-kochetov (approved):

PR #2577, opened 2024-10-08
Bugfix/2532 search result is not opening in mobile view if chat is the same as active one

frolyo:


src/dotnet/UI.Blazor.App/Services/Search/SearchUI.cs
```diff
@@ -97,15 +100,25 @@ private void NavbarUIOnSelectedGroupChanged(object? sender, NavbarGroupChangedEv
 
     public void Select(FoundItem foundItem)
     {
```
crui3er, line 102: Why SelectPrevious/SelectNext are implemented with using NavigateTo, but this method has completely different implemention?

frolyo: good question. In fact it's not needed since Select is called from navbar and already does navigation.
But I can call Navigate from Select for code consistency between the method.

PR #2721, opened 2024-12-26
Feat/2687 chat roulette

crui3er:


src/dotnet/Api/Contacts/Contact.cs
```diff
@@ -20,6 +20,7 @@ public sealed partial record Contact(
     [DataMember, MemoryPackOrder(3)] public Moment TouchedAt { get; init; }
     [DataMember, MemoryPackOrder(4)] public bool IsPinned { get; init; }
     [DataMember, MemoryPackOrder(7)] public string PeerContactName { get; init; } = "";
+    [DataMember, MemoryPackOrder(8)] public bool IsChatRoulette { get; init; }
```
alexis-kochetov: Doesn't look well, let's try to get rid of it

alexis-kochetov: Probably you can determine that by Roulette placeId

alexis-kochetov (changes requested):

src/dotnet/Api/Contacts/Contact.cs
crui3er: Reworked with using SystemTag property

alexis-kochetov (approved):

PR #2766, opened 2025-01-27
fix: user unexpectedly leaves place on chat leaving #2762

frolyo:


alexis-kochetov (approved):

tests/Chat.IntegrationTests/ChatOperationsTest.cs
```diff
@@ -353,12 +353,12 @@ public async Task PromoteAuthorToOwner(bool isPublicChat)
     {
         var appHost = AppHost;
         await using var ownerTester = appHost.NewBlazorTester(Out);
-        await ownerTester.SignInAsAlice();
+        await ownerTester.SignInAsUniqueAlice();
```
crui3er, line 356: Why did you renamed to SignInAsUniqueXXX? To prevent tests affects each other's data due to same host and same users are used?

crui3er (approved):

PR #2805, opened 2025-02-07
Fix/update UI 86

iqmulator:


src/dotnet/UI.Blazor/Components/Form/Toggle.razor
```diff
@@ -27,6 +33,19 @@
 
     public bool CurrentValue { get; set; }
 
+    protected override void OnAfterRender(bool firstRender) {
+        if (firstRender) {
```
crui3er: I suppose it's to late to make vibration.
I would expect vibration feedback just after I clicked a toggle (or panel with toggle).
Here it's much later.
Why did not implement it in OnParametersSet?

src/dotnet/UI.Blazor/Components/Form/Toggle.razor
```diff
@@ -27,8 +33,18 @@
...
 
-    protected override void OnParametersSet()
-        => CurrentValue = IsChecked;
+    protected override void OnParametersSet() {
+        if (_previousValue is null) {
+            _previousValue = IsChecked;
+        } else {
+            if (_previousValue.Value != IsChecked) {
+                _previousValue = IsChecked;
+                _ = TuneUI.Play(Tune.ChangeToggle);
```
frolyo, line 42: Vibration is supposed to be on user action, not on parameter set

frolyo (changes requested):

frolyo (approved):

PR #2887, opened 2025-03-20
fix: image-skeleton for video, VisualMediaViewer, video thumbnail, css

iqmulator:


src/dotnet/UI.Blazor/Components/VideoAttachmentThumbnail/VideoAttachmentThumbnail.razor
```diff
@@ -22,13 +24,15 @@
     [Parameter] public TextEntryAttachment Attachment { get; set; } = null!;
     [Parameter] public string Class { get; set; } = "";
 
-    private string GetUrl()
+    private (string, string) GetUrls()
```
alexyakunin: You should use `(string OriginalUrl, string ThumbnailUrl)` here, coz it's unclear what they are.

alexyakunin (changes requested):

PR #2951, opened 2025-04-25
fix: chat-view and panels width, translation btn, banners, record-btn, etc

iqmulator:


src/nodejs/styles/main.css
```diff
@@ -193,9 +194,17 @@ body.narrow .layout-header > .c-content.selection-header {
 }
 .list-view-layout > .layout-header > .c-content {
     @apply flex-1;
-    @apply md:max-w-196;
+    @apply lg:max-w-196;
     @apply md:p-2.5;
 }
+/* Chat View in other screen sizes */
```
crui3er: Empty selectors. Should we remove them?

crui3er (approved):

src/nodejs/styles/main.css
iqmulator: Removed. Thanks.

PR #2957, opened 2025-04-29
fix: modals, chat title, side panels, update: editor, add: new icons, refactor: ui

iqmulator:


src/dotnet/UI.Blazor.App/Components/ChatMessageEditor/chat-message-editor.ts
```diff
@@ -119,6 +123,19 @@ export class ChatMessageEditor {
 
     }
 
+    private updatePostPanelBorderRadius = (entries) => {
+        let clsLst = this.postPanelDiv.classList;
```
crui3er, line 127: if variable is not supposed to be reassigned later, use const keyword

crui3er (approved):

PR #2959, opened 2025-05-05
Fix/update UI 113

iqmulator:


src/dotnet/UI.Blazor.App/Components/ChatAudioPanel/RecorderToggle.razor
```diff
@@ -127,9 +127,10 @@
         await JS.InvokeVoidAsync(JSInitMethod);
     }
 
-    public override ValueTask DisposeAsync() {
+    public override async ValueTask DisposeAsync() {
         _animator.Dispose();
-        return base.DisposeAsync();
+        await JS.InvokeVoidAsync($"{BlazorUIAppModule.ImportName}.RecorderToggle.dispose");
```
crui3er, line 132: This line will fail with an exception if client has disconnected in SSB mode.
Wrap this call with try/catch

src/dotnet/UI.Blazor.App/Components/ChatAudioPanel/recorder-toggle.ts
```diff
@@ -2,17 +2,73 @@
...
 
     public static async init(): Promise<void> {
         if (this.isInitialized)
             return;
 
         const buttons = [...document.querySelectorAll<HTMLButtonElement>('div.recorder-wrapper > button')];
         buttons.forEach(btn => {
-            btn.addEventListener('click', () => recordingAudioContextSource.initContextInteractively());
-            btn.addEventListener('click', () => audioContextSource.initContextInteractively());
+            const handler1 = () => recordingAudioContextSource.initContextInteractively();
```
crui3er, line 14: Why do we need 2 handles and 2 subscriptions here?
I think it would better to combine them:
const handler = () => {
   recordingAudioContextSource.initContextInteractively();
   audioContextSource.initContextInteractively();
}

src/dotnet/UI.Blazor.App/Components/ChatMessageEditor/chat-message-editor.ts
```diff
@@ -148,6 +149,10 @@ export class ChatMessageEditor {
                     fromEvent(this.attachmentListElement, 'wheel')
                         .pipe(takeUntil(this.disposed$))
                         .subscribe((event: WheelEvent) => this.onHorizontalScroll(event));
+                } else if (element.className == 'related-chat-entry-panel') {
+                    if (!this.editorDiv.classList.contains('with-related-panel')) {
```
crui3er, line 153: Do we really have any benefits by managing these classes manually rather then just using 'has' css selector?

crui3er (changes requested):

PR #3073, opened 2025-06-23
feat: support og tags for link to text entry

crui3er:


src/dotnet/Chat.Service/ContentLinksBackend.cs
```diff
@@ -5,22 +5,20 @@ namespace ActualChat.Chat;
...
 {
-    [field: AllowNull, MaybeNull]
-    private IAccountsBackend AccountsBackend => field ??= services.GetRequiredService<IAccountsBackend>();
-    [field: AllowNull, MaybeNull]
-    private IChatsBackend ChatsBackend => field ??= services.GetRequiredService<IChatsBackend>();
-    [field: AllowNull, MaybeNull]
-    private IAuthorsBackend AuthorsBackend => field ??= services.GetRequiredService<IAuthorsBackend>();
-    [field: AllowNull, MaybeNull]
-    private IPlacesBackend PlacesBackend => field ??= services.GetRequiredService<IPlacesBackend>();
+    private IAccountsBackend AccountsBackend { get; } = services.GetRequiredService<IAccountsBackend>();
```
alexis-kochetov, line 8: What is the reason to switch lazy init for dependencies?

crui3er: We normally use lazy initialization to prevents conflicts when services can resolve each other.
Here it is not required. So I decided to put it to normal initialization.

alexis-kochetov (approved):

PR #3128, opened 2025-07-25
fix: animation for translating

iqmulator:


src/dotnet/UI.Blazor.App/Components/ChatView/Items/ChatEntryMessageInternalView.razor
```diff
@@ -185,6 +196,8 @@
         var isRenderedTranslated = (_rendered?.Translation ?? TranslationState.None) != TranslationState.None;
         var isRenderedStreaming = _rendered?.IsStreaming ?? false;
         var entry = await ChatUI.GetEntry(renderedEntry.Id, cancellationToken).ConfigureAwait(false);
+        var oldState = _rendered;
+        _oldTranslation = _rendered?.Translation ?? TranslationState.None;
```
crui3er: `_oldTranslation` should be calculated in Render part (top of the file with markup).
We should avoid modifying class fields from the `ComputeState` method. It can return result only via Model.

crui3er (changes requested):

crui3er (approved):

PR #3129, opened 2025-07-25
fix: ConversationMessageView, update VisualMediaViewer

iqmulator:


src/dotnet/UI.Blazor/Services/VisualMediaViewerUI.cs
```diff
@@ -5,9 +5,9 @@ public class VisualMediaViewerUI(UIHub hub)
 {
     private ModalUI ModalUI => hub.ModalUI;
 
-    public async Task Show(string url, ChatEntry? chatEntry = null)
+    public async Task Show(string url, Conversation? conversation = null, ChatEntry? chatEntry = null)
```
crui3er: I think it would be better to pass an array of attachments instead of a ChatEntry or a Conversation.
The only issue i see now how to handle isYouTubeVideo check in Viewer. But to be honest, i don't understand when it can be true and why it uses link previews. May be on mobile? It should be checked or consulted with the team.

crui3er (changes requested):

src/dotnet/UI.Blazor.App/Components/ChatView/Items/Attachment/VisualMediaAttachmentList.razor
```diff
@@ -10,29 +10,28 @@
...
+    [Parameter, EditorRequired] public List<TextEntryAttachment> MediaList { get; set; } = null!;
 
     private IEnumerable<ImageRow> GetRows() {
         var line = new List<ImageTile>();
-        foreach (var attachment in ImageList) {
+        foreach (var attachment in MediaList) {
             var tile = new ImageTile(attachment);
             if (attachment.Index == 1
-                && attachment.Index == ImageList.Count - 1
+                && attachment.Index == MediaList.Count - 1
```
crui3er: It's incorrect to use attachment.Index here. This property shows index of attachment inside ChatEntry. Apparently here you need to use index of the item inside MediaList

src/dotnet/UI.Blazor/Components/VisualMediaViewerModal/VisualMediaViewerModal.razor
```diff
@@ -152,9 +153,16 @@
         BlazorRef = null!;
     }
 
-    protected override void OnInitialized() {
-        _attachments = ModalModel.ChatEntry?.Attachments.ToList() ?? new ();
-        var selectedAttachment = _attachments.FirstOrDefault(x => OrdinalIgnoreCaseEquals(UrlMapper.ContentUrl(x.Media.ContentId), ModalModel.Url));
+    protected override async Task OnInitializedAsync() {
+        var attachmentTasks = ModalModel.Attachments
+            .Select(async textEntryAttachment => {
+                var chatEntry = await Chats.GetEntry(Session, textEntryAttachment.EntryId, CancellationToken.None).ConfigureAwait(false);
```
crui3er: It's not needed to fetch chatEntry, ChatEntry is used only to get link previews and check whether you tube video player is needed. But this seems no longer needed. VisualMediaViewer is used only for displaying local link previews.
Moreover, now it works really strange (in fact, it's a bug).
If we have a chat entry with image attached and link to youtube video. Viewer will show image and will paly video in the same time.

crui3er (changes requested):

src/dotnet/UI.Blazor.App/Components/ChatView/Items/Attachment/VisualMediaAttachmentList.razor
```diff
@@ -10,37 +10,43 @@
...
+            MediaList="@MediaList"/>
     }
 </div>
 
 @code {
     private IEnumerable<ImageRow> _rows = null!;
 
-    [Parameter, EditorRequired] public List<TextEntryAttachment> ImageList { get; set; } = null!;
-    [Parameter] public ChatEntry? ChatEntry { get; set; }
+    [Parameter, EditorRequired] public List<TextEntryAttachment> MediaList { get; set; } = null!;
```
crui3er: Use TextEntryAttachment[] type

crui3er (approved):

PR #3130, opened 2025-07-28
fix: conversation misses some attachments

crui3er:


src/dotnet/Chat.Service/ConversationsBackend.cs
```diff
@@ -390,7 +390,8 @@ private async Task<ConversationEntriesInfo> GetTextEntries(ChatId chatId, Range<
             .Where(e => entryIdRanges.Any(r => r.Contains(e.LocalId)))
             .ToArray();
         foreach (var entry in chatEntries) {
-            textEntries.Add(new TextEntry(entry));
+            if (!entry.Content.IsNullOrEmpty())
```
alexis-kochetov: It looks suspicious - probably you may lose attachments again

alexis-kochetov (approved):

PR #3273, opened 2025-10-14
Add offline message persistence and delayed file upload handling

crui3er:
This PR introduces several improvements to the message sending and file upload logic to enhance reliability and offline support:

New features

Offline message persistence:
Outgoing messages are now stored locally when there is no connection to the server. They are automatically sent once the connection is restored — even after the app has been restarted.

Deferred file uploads:
Files are now uploaded after the user presses the “Send” button, instead of being uploaded in advance.

Resumable uploads:
If the connection to the server is lost or the app is restarted during a file upload, the process will automatically resume when the connection becomes available again.

Additional notes

Improves reliability for users with unstable or intermittent network connections.

Ensures consistent message and file delivery after app restarts or network interruptions.


src/dotnet/Core/Kvas/IKvas.cs
```diff
@@ -9,3 +9,8 @@ public interface IKvas : IHasServices
 }
 
 public interface IKvas<TScope> : IKvas;
+
+public interface IKvas2 : IKvas
```
alexis-kochetov: Are you going to keep this name?

alexis-kochetov (approved):

src/dotnet/Core/Kvas/IKvas.cs
crui3er: I got rid of IKvas2. Only BatchingKvas implements GetAll.

PR #3277, opened 2025-10-15
chore: fix warnings

crui3er:


src/dotnet/UI.Blazor.App/Components/IncomingShare/ReadonlyAttachmentList.cs
```diff
@@ -5,7 +5,7 @@ public sealed class ReadonlyAttachmentList : IAttachmentList
     private readonly List<Attachment> _attachments;
     public int Count => _attachments.Count;
     public IEnumerable<Attachment> Items => _attachments;
-    public event EventHandler? Changed;
+    public event EventHandler? Changed { add  {} remove { } }
```
frolyo, line 8: looks suspicious. why would we need fake event doing nothing?

alexis-kochetov (approved):

src/dotnet/UI.Blazor.App/Components/IncomingShare/ReadonlyAttachmentList.cs
crui3er: This event is a part of IAttachmentList interface. This implementation never invokes this event.

PR #3282, opened 2025-10-16
fix: selection-header, buttons

iqmulator:


src/dotnet/UI.Blazor/Components/Button/button.css
```diff
@@ -285,25 +279,33 @@ body.hoverable .btn.btn-round.btn-transparent:hover,
...
+.touch-capable .btn.btn-square.btn-primary:active,
+body.hoverable .btn.btn-round.btn-primary:hover,
+.touch-capable .btn.btn-round.btn-primary:active {
     @apply text-primary;
 }
 
 /* Button Square and Round Danger */
 body.hoverable .btn.btn-square.btn-danger:hover::after,
-body.hoverable .btn.btn-round.btn-danger:hover::after {
+.touch-capable .btn.btn-square.btn-danger:active::after,
```
crui3er, line 299: Why do you need some many repeating selectors?

src/dotnet/UI.Blazor/Components/Button/button.css
```diff
@@ -285,25 +279,33 @@ body.hoverable .btn.btn-round.btn-transparent:hover,
...
+body.hoverable .btn.btn-round.btn-danger:hover::after,
+.touch-capable .btn.btn-round.btn-danger:active::after {
```
crui3er, line 301: Why do you use `body.hoverable` with body prefix and `.touch-capable` without body prefix while both are defined on body?

crui3er: Just looking for a solution how can we simplify button css and make it more maintainable.

crui3er (approved):

PR #3292, opened 2025-10-16
chore: streamline app base disposing

crui3er:


alexyakunin (commented):
Does it make sense to use standard dispose pattern here?

PR #3301, opened 2025-10-25
fix: chat-entry-message-internal-view.ts

iqmulator:


src/dotnet/UI.Blazor.App/Components/ChatView/Items/ChatEntryMessageInternalView/chat-entry-message-internal-view.ts
```diff
@@ -1,6 +1,13 @@
 import { Subject } from 'rxjs';
 import { debounce, ResettableFunc } from 'promises';
 
+const importantClasses = new Set([
```
crui3er: may be call 'steamingMessageClasses' ?

PR #3362, opened 2025-11-26
fix: subheaders are lit-components with animation

iqmulator:


src/dotnet/UI.Blazor.App/Components/SubHeader/sub-header.lit.ts
```diff
@@ -0,0 +1,65 @@
+import {customElement, property} from "lit/decorators.js";
+import {css, html, LitElement} from "lit";
+
+@customElement('sub-header')
+class SubHeaderSvg extends LitElement {
```
crui3er: Why does it have suffix Svg?
What is the purpose of the component?

src/dotnet/UI.Blazor.App/Components/SubHeader/SubHeader.razor
```diff
@@ -0,0 +1,68 @@
...
+        BlazorRef.DisposeSilently();
+        BlazorRef = null!;
+    }
+
+    protected override async Task OnAfterRenderAsync(bool firstRender) {
+        if (!_render)
+            return;
+
+        if (!_jsInitialized) {
+            _jsInitialized = true;
```
crui3er, line 38: `_jsInitialized` is the same as BlazorRef != null

src/dotnet/UI.Blazor.App/Components/SubHeader/sub-header.ts
```diff
@@ -0,0 +1,47 @@
...
+    static create(subHeaderLit: HTMLElement, blazorRef: DotNet.DotNetObject): SubHeader {
+        return new SubHeader(subHeaderLit, blazorRef);
+    }
+
+    constructor(
+        private readonly subHeaderLit: HTMLElement | null,
+        private readonly blazorRef: DotNet.DotNetObject
+    ) {
+        this.subHeaderLit = subHeaderLit;
+        if (this.subHeaderLit == null)
```
crui3er: When can it be null?

src/dotnet/UI.Blazor.App/Components/SubHeader/sub-header.ts
```diff
@@ -0,0 +1,47 @@
...
+        if (this.disposed$.closed)
+            return;
+
+        this.disposed$.next();
+        this.disposed$.complete();
+
+    }
+
+    public setHeight() {
+        if (this.subHeaderLit == null)
```
crui3er: Why do you need this check?

src/dotnet/UI.Blazor.App/Components/SubHeader/SubHeader.razor
```diff
@@ -0,0 +1,68 @@
+@using ActualChat.UI.Blazor.App.Module
+@namespace ActualChat.UI.Blazor.App.Components
+@inherits ComponentBase<AppUIHub>
+
+@if (_render) {
```
crui3er, line 5: Looks like it possible that we might the following lifecycle:
_render: false, true, false, true (again)
Can component handle this case properly?
I guess might be issues if _render is turned true due to setting IsVisible true again while OnAnimationEnd is not yet invoked. In that case we will have new sub-header rendered while js still refers to old one.

crui3er (changes requested):

PR #3401, opened 2025-12-22
feat: hide recaptcha error message for user

iqmulator:


src/dotnet/UI.Blazor.App/recaptcha-handler.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    public static init() {
+        this.observer = new MutationObserver((mutations) => {
+            for (const mutation of mutations) {
+                for (const node of Array.from(mutation.addedNodes)) {
+                    if (node.nodeType !== Node.ELEMENT_NODE)
+                        continue;
+
+                    const outerDiv = node as HTMLElement;
+                    if (outerDiv.parentElement !== document.body)
```
crui3er: Can it be a different parent?

iqmulator: no, i removed it

PR #3512, opened 2026-02-04
Fix/update UI 203

iqmulator:


frolyo (approved):

src/dotnet/UI.Blazor/Components/VirtualList/virtual-list.ts
```diff
@@ -1201,14 +1200,17 @@ export class VirtualList {
...
 
-                let hasAnchor = false;
-                fastRaf({
-                    read: () => { hasAnchor = lastItemRef.classList.contains('anchor'); },
-                    write: () => {
-                        if (!hasAnchor)
-                            lastItemRef.classList.add('anchor');
-                    },
-                });
+                const hasAnchor = lastItemRef.classList.contains('anchor');
```
alexis-kochetov: You should not read anything in DOM during write phase of the fast RAF abstract. So you need to separate logic - read during the read phase and then perform any writes during write phase

alexis-kochetov (changes requested):

src/dotnet/UI.Blazor/Components/VirtualList/virtual-list.ts
iqmulator: done

alexis-kochetov (approved):

PR #3522, opened 2026-02-05
fix: improve keyboard handling on Android, add platform-specific styles

iqmulator:


src/dotnet/App.Maui/MainPage.cs
```diff
@@ -22,6 +22,8 @@ public MainPage()
 #endif
         BackgroundColor = MauiSettings.SplashBackgroundColor;
         RecreateWebView();
+
+        Loaded += OnLoaded;
```
frolyo, line 26: unsubscrbing is not done

frolyo (changes requested):

src/dotnet/App.Maui/MainPage.cs
iqmulator: done

frolyo (approved):

PR #3604, opened 2026-02-27
feat: route to online nodes only

alexis-kochetov:


alexyakunin:
I made a better fix: there is no Offline state now. Discarding, but thanks for identifying the issue.

PR #3615, opened 2026-03-05
feat: Better connectivity handling

alexyakunin:


src/dotnet/App.Maui/Services/NetworkAccessExt.cs
```diff
@@ -0,0 +1,7 @@
+namespace ActualChat.App.Maui.Services;
+
+public static class NetworkAccessExt
+{
+    public static bool IsOnline(this NetworkAccess networkAccess)
```
crui3er, line 5: Why did you not exclude NetworkAccess.ConstrainedInternet. In this case i suppose it'not possible to connect to voxt.ai.

PR #3629, opened 2026-03-11
Feat/1xxx crash on android 9

crui3er:
- fixed app crash on Android 9
- fixed ugly app appearance due to missing safe-area-inset-* css environment vars


src/dotnet/App.Maui/MauiThemeHandler.cs
```diff
@@ -94,7 +94,15 @@ protected virtual bool Apply(string topBarColor, string bottomBarColor, Theme? t
             Theme.Dark => StatusBarStyle.LightContent,
             _ => StatusBarStyle.Default,
         };
+
+#if ANDROID
```
frolyo, line 98: how about using
```csharp
if (OperatingSystem.IsAndroidVersionAtLeast(...
```

frolyo (approved):

PR #3678, opened 2026-03-30
fix: SettingsPanel, LanguageButtonGroup, JoinVideoCallModal, search-input

iqmulator:


src/dotnet/UI.Blazor/Components/SettingsPanel/SettingsPanel.razor
```diff
@@ -134,7 +134,11 @@
     protected override void OnAfterRender(bool firstRender) {
         var currentTabs = _tabRegistry.SortedTabs.ToList();
         var selectedTabId = SelectedTabId;
-        SelectedTabId = _tabRegistry.ResolveSelectedTabId(currentTabs, _renderedTabs, _renderedSelectedTab);
+        // On first render, tabs register during ChildContent (after _renderedTabs snapshot),
+        // so _renderedSelectedTab is null even though SelectedTabId may be valid.
+        // Skip resolution when the current selection is already valid in the new tab set.
+        if (_renderedSelectedTab != null || selectedTabId == null || currentTabs.All(t => t.Id != selectedTabId))
```
crui3er: Looks complicated. I think we should just add to the beginning of the method:
```
if (firstRendere) {
   StateHasChanged();
   return;
}
```
On the first render it registers Tab. On the second it will render selected tab content (_renderedSelectedTab  will receive its value) and since this moment we can invoke ResolveSelectedTabId.
What do you think?

Also you need to check registry usage in TabPanel component. I suppose there might be similar issue.

crui3er (changes requested):

src/dotnet/UI.Blazor/Components/SettingsPanel/SettingsPanel.razor
iqmulator: done

crui3er (approved):

PR #3690, opened 2026-03-31
fix: support legacy emojis in reaction lookup

iqmulator:


src/dotnet/Api/Identifiers/Emojis.cs
```diff
@@ -76,6 +86,16 @@ public static class Emojis
     /// </summary>
     public static readonly Emoji[] Legacy = [
```
crui3er, line 87: Should we make Legacy Emoji list and legacy emoji fields private?

crui3er (approved):

PR #3699, opened 2026-04-01
fix: 'A task was cancelled' error toast is shown periodically when there is a sending message and no connection

crui3er:


src/dotnet/UI.Blazor.App/Services/SendingMessages/SendingMessages.Queue.cs
```diff
@@ -23,9 +23,13 @@ partial class SendingMessages
                         command.Request.Text.ToPrivate());
                     throw;
                 }
-                Log.LogWarning(e,
-                    "ProcessQueueItem failed for '{Text}', retrying in {Delay}s",
-                    command.Request.Text.ToPrivate(), ProcessCommandRetryDelay.TotalSeconds);
+                if (e is OperationCanceledException)
```
alexyakunin, line 26: `e.IsCancellationOf(ct)` is a better check, coz OpCancelledException might originate from some other token.

crui3er: I think it's already covered with `catch (Exception e) when (!cancellationToken.IsCancellationRequested) {` above

PR #3780, opened 2026-04-21
misc: Video capture/preview features that were overlapping with @crui3er commits (prob we won't merge that)

alexyakunin:
What's non-overlapping is already in dev.


alexyakunin:
I migrated whatever makes sense. Closing w/o merge.

PR #3892, opened 2026-05-21
feat: New INotifications/NotificationsBackend, NotificationItem union, etc.

alexyakunin:
… contracts

Phase 1 of the notifications redesign (docs/plans/notif-api.md):
- NotificationItem: abstract-record union (ChatNotificationItem, AttentionNotificationItem), MessagePack-only, base carries the identity/dedup key and the ReadEntryLid read-detection anchor
- UserNotificationInfo / NotificationDelta: per-user notification state


alexis-kochetov:
Addressed the review comments in 6b23124fd:

- **`notif-api.md` §3.3 data model** — updated to match the shipped code: `Notification` / `ChatNotification` (not `NotificationItem` / `ChatNotificationItem`), `ChatId` is **derived** (not stored), and documented `ChatEntryRelatedNotification` vs `ChatEntryNotification` plus the real per-kind type names incl. `ThreadNotification` (was `NewThreadNotificationItem`).
- **`notif-api.md` status line + Phase 1** — client-side reconciliation is shipped (prune + create on web/Android, iOS prune-only; see §8), not deferred.
- **`notification-ui.ts` `reconcileNotifications`** — switched from `getRegistration('sw.js')` to `navigator.serviceWorker.ready`, which reliably resolves to the active registration for our scope (the URL lookup can return `undefined` even when the SW is registered).

`build:Verify` green.

PR #3996, opened 2026-07-02
fix: overlay full-screen avatar; drop Block user toggle

iqmulator:


src/dotnet/UI.Blazor.App/Components/RightPanel/right-panel-header.ts
```diff
@@ -12,6 +12,7 @@ import { fastRaf } from 'fast-raf';
 export class RightPanelHeader {
     private blazorRef: DotNet.DotNetObject;
     private readonly header: HTMLElement;
+    private readonly varTarget: HTMLElement;
```
crui3er, line 15: why is not just 'target'?

crui3er (approved):

PR #4049, opened 2026-07-13
fix(db): make _events inserts idempotent via ON CONFLICT DO NOTHING

alexis-kochetov:
Deterministic-Uuid operation events (delay-quantized FlowResumeEvent, KeyConflictStrategy.Skip) let concurrent producers race on the _events primary key, surfacing as recurring `duplicate key value violates unique constraint "pk_events"` errors that fail the enclosing command (~5/day in prod).

Annotate DbEvent with ConflictStrategy.DoNothing in the shared DefineIndexes helper so every operational DbContext emits `INSERT ... ON CONFLICT DO NOTHING`, turning the race into an idempotent no-op. It must be set on all contexts, not just Flows: NpgsqlUpdateSqlGenerator caches the strategy in a static dictionary keyed by CLR type, and DbEvent is mapped in all 8 contexts, so the first inserter would otherwise cache Unspecified globally.

Adds annotation-only (no-DDL) migrations for all 8 contexts and an integration test asserting the generated SQL carries ON CONFLICT DO NOTHING and that a duplicate insert no longer raises 23505.


alexyakunin:
Notes from Claude:

That PR is a great pointer — and it validates your instinct: the app-side blanket fix is subtly wrong, which is exactly why Fusion should own this. Let me lay it out.

## What's actually happening

`DbOperationScope.Commit` resolves `KeyConflictStrategy.Skip`/`Update` events via **check-then-insert**: `FindAsync` → not found → `Add` (`DbOperationScope.cs:242-257`). Two concurrent producers of a deterministic-UUID event (delay-quantized `FlowResumeEvent`) both find nothing, both insert, one hits `23505` at commit → the **entire command transaction fails** (operation row included), gets classified transient, and `OperationReprocessor` re-executes the whole command. It self-heals, but at the cost of: an error log per race, a full command re-execution (side effects run twice!), and added latency. That's the ~5/day noise in prod.

## Why the actual-chat fix isn't the right shape for Fusion

Annotating `DbEvent` with a blanket `ON CONFLICT DO NOTHING` makes *every* insert idempotent — which **breaks the other two strategies**:

- `KeyConflictStrategy.Fail` must fail loudly on conflict — DO NOTHING silently swallows it.
- `KeyConflictStrategy.Update`'s insert-when-absent leg silently no-ops instead of updating when a row appears concurrently.
- Plus the EF wrinkle their custom generator has to paper over: DO NOTHING returns 0 affected rows, which EF normally reports as `DbUpdateConcurrencyException`.

Fine as an app-side stopgap where all events are Skip-semantics; wrong as framework behavior. Your "handle all values correctly" constraint rules it out.

## Proposed approaches

**Preferred: savepoint-based conflict resolution loop in `Commit` (provider-agnostic, correct for all strategies).** Split the events flush from the operation-row flush (same transaction). When any event has strategy ≠ `Fail`: create a savepoint → `SaveChanges` (events) → on unique violation (or version-conflict for `Update`): roll back to the savepoint, re-`FindAsync` the conflicting UUIDs, apply the strategy in memory — `Skip` → drop the add; `Update` → `State == New` check then update, else the existing `KeyConflictResolver.Error`; `Fail` never enters the loop — retry the save. Log at **Debug**, since a resolved conflict is normal operation for deterministic-UUID events.

- Correct per-strategy semantics, no schema/migrations, works on every provider (EF savepoints; PG needs them because `23505` dooms the transaction, SQL Server/MySQL tolerate them anyway).
- Extra round trips only on an actual conflict (~5/day scale — negligible).
- Kills the command-level retry, so no double side effects and no error noise.
- Bonus: it composes with the item-11-follow-up stable-UUID work — that unique-violation-as-signal path needs similar handling anyway.

**Option B: provider upsert seam.** An interface (e.g. event-insert strategy resolved per provider) where `ActualLab.Fusion.EntityFramework.Npgsql` emits real `INSERT ... ON CONFLICT DO NOTHING / DO UPDATE ... WHERE state = New` per strategy, with the savepoint loop as the portable fallback. Zero-race and one round trip, but per-provider SQL generation, affected-rows handling, and a bigger surface. Worth it later as an optimization if conflict volume ever grows; overkill for ~5/day.

**Option C: entity-level `ON CONFLICT DO NOTHING` (the PR's approach) in Fusion.** Rejected: breaks `Fail`/`Update` semantics as above.

**Option D: serialize producers via advisory locks on the event UUID.** Provider-specific, adds a round trip to *every* event insert to solve a rare race. Rejected.

My recommendation: **A now**, with B noted as a deferred optimization — and once it ships, actual-chat can drop the annotation + 8 migrations entirely.

--

Overall, I'll take care of that in Fusion, don't merge.

alexyakunin:
Overall, I'll take care of that in Fusion, don't merge.

PR #4092, opened 2026-07-30
feat(chat): notification left panel — Unread group with All/DMs/Mentions tabs

iqmulator:
Add a bell in the left rail (below the roster, above the avatar) that selects a new "Unread" navbar group rendering a cross-place notifications panel with All / DMs / Mentions tabs (NotificationsNavbarWidget + per-tab unread badges).

- ChatListFilter: AcrossPlace flag + Unread/UnreadDMs/UnreadMentions filters; ChatListUI.ListUnordered routes AcrossPlace filters through ListAllUnordered.
- Mentions.GetLastOwn now considers both user-ref (u:) and author-ref (a:) mentions and keeps the most recent; ChatInfo.HasUnreadOwnMention drives the chat-list "@" badge.
- Bell is hidden when there are no unread chats; "Back" from a chat opened via the Unread panel returns to it while unread remain, else falls back to Chats (invariant: Unread group selected iff unread chats exist).
- Plus moved off the rail into a floating FAB (ChatListCreateButton).
- New notify-bell icon (font regenerated).


src/dotnet/Chat.Service/Mentions.cs
```diff
@@ -21,6 +21,22 @@ public class Mentions(IServiceProvider services) : DbServiceBase<ChatDbContext>(
...
+        var authorMentionTask = Backend.GetLast(chatId, MentionRef.NewAuthor(author.Id), cancellationToken);
+        var userMentionTask = author.UserId.Value.Length != 0
+            ? Backend.GetLast(chatId, MentionRef.NewUser(author.UserId), cancellationToken)
+            : null;
+
+        var authorMention = await authorMentionTask.ConfigureAwait(false);
+        var userMention = userMentionTask != null
+            ? await userMentionTask.ConfigureAwait(false)
+            : null;
+        return Latest(authorMention, userMention);
```
alexyakunin, line 35: AFAIK it inserts user mention today in almost any case except in anon chats, where it inserts author mention. so assuming we don't care about too old mentions, the logic here should be _either_ author or user (based on chat type), but not both, I guess. more precisely, we can have "has at least 1 author mention" flag for chat - it can be even stored, ~ equivalent to "the chat was anon at some point in past". and if there is such a flag, we should search both user and author mentions; otherwise it's ok to search just user mentions.

alexyakunin (approved):
Approve, but let's keep this in mind and maybe fix later. It's ok to have some inefficiency on backend for now.

PR #4097, opened 2026-07-31
feat(chat-list): keep read chats in the unread panel until it's closed

iqmulator:


alexyakunin:
I don't like it. "Unread" is an order, not "Panel".

PR #4106, opened 2026-08-03
feat(chat): add Moderator role

alexyakunin:
A third chat system role between `Anyone` and `Owner`.

Moderators can delete and restore **anyone's** messages, mute participants and manage live sessions, remove members, and edit chat title/picture/description — but **not** delete or archive the chat, change its type, or edit roles.

Plan: [`docs/plans/moderator-role.md`](docs/plans/moderator-role.md)

## Capability matrix

| Capability | Anyone | Moderator | Owner |
|---|:--:|:--:|:--:|
| Delete / restore anyone's messages | ❌ | ✅¹ | ✅ |
| Mute a participant, mute all, manage session rules | ❌ | ✅¹ | ✅ |
| Remove (kick) a member | ❌ | ✅¹ | ✅ |
| Edit chat title, picture, description | ❌ | ✅ | ✅ |
| Change chat type, alias, guest/anon policy | ❌ | ❌ | ✅ |
| Archive / delete chat, edit roles, reassign call host | ❌ | ❌ | ✅ |

¹ **Owners are immune.** A Moderator who isn't an Owner can't delete an Owner's message, mute an Owner, or kick one. Immunity resolves through a new *unmasked* backend helper rather than `IRoles.ListOwnerIds` — that one masks anonymous owners and would have silently stripped their immunity.

## Call authority

Three independent sources; the actor holds the **union**:

- **Owner** — may mute anyone, including other Owners and the host.
- **Call host** — may mute anyone, Owners included. The host runs *this call*; a Moderator polices the *chat*.
- **Moderator** — may mute anyone **except** an Owner.

Because host ⇒ Owner-muting, **`SetHost` (new) is an Owner-or-current-host power only** — otherwise a Moderator could grant themselves that power and Owner immunity would be vacuous. This is the one place the implementation is stricter than originally specced; it has a dedicated test, as does the union rule itself.

The host slot also reassigns when its holder leaves (preferring a remaining Owner), so a call can't be stranded with an absent host nobody can replace. `MuteAll` widens to an exception set so a Moderator's mute-all leaves Owners unmuted.

## Place-level moderators

Mostly free: place permissions *are* the root chat's, and public place chats inherit root-chat rules verbatim. Private place chats need their own appointment — same as for Owners today.

## API change: `Authors_PromoteToOwner` → `Authors_ChangeRole`

Replaced with the general `Authors_ChangeRole(Session, AuthorId, SystemRole, bool IsInRole)` (and `Places_ChangeRole`). The old commands stay one release as `[Obsolete]` shims that **rewrite into** the new one rather than duplicating its logic — `IApiCommand : IDelegatingCommand`, so dispatching is precisely what these handlers are contractually for, and the nested command still runs the full pipeline with its own permission checks. A regression test asserts the shim *inherits* the new behaviour.

Owner demotion (`Owner, false`) is rejected for now, preserving today's semantics; the command shape makes enabling it later a one-line change.

## Notes for review

- **Migration snapshot churn.** `ChatDbContextModelSnapshot.cs` reformats heavily — EF-11 codegen style, since the checked-in file came from EF 10.0.8 tooling while the project is on `11.0.0-preview.6`. Diffed whitespace-normalized, the only model change is `CanModerate`; the migration itself is a single `AddColumn`. Will conflict with any concurrent migration branch.
- **Client/server skew is safe.** `Moderate` is a new flag on an existing enum in an existing payload — old clients ignore it, and the legacy commands still work.
- **Archived chats:** `GetConsolidatedRules` returns `None` for non-Owners, so Moderators lose access there. Consistent with "Moderators can't archive"; left as-is.

## Testing

| Suite | Result |
|---|---|
| `ActualChat.CI.slnf` build | clean |
| `Chat.UnitTests` | 458 passed |
| `Chat.IntegrationTests` (full) | 359 passed |
| new `ModeratorRoleTest` | 10 passed |
| new `CallModerationTest` | 11 passed |

`Streaming.IntegrationTests` (slow real-audio suite) was not run to completion — no test in it references `MuteAll`, `SetHost`, or `ILiveSessions`, and it compiles as part of the clean CI build.

Two bugs the new tests caught during development: `SetHost` invalidated `Get` instead of `GetState` (state write, stale read), and `LiveSessionsBackend.Get` returns null for ambient sessions so the first test arrange had no members to observe.

https://claude.ai/code/session_015WG3e7dyH2DUMDTNV3TKxT


alexyakunin:
## Follow-up: a hole the audit caught

**`Authors.OnExclude` let a Moderator kick an anonymous Owner.** It resolved the owner list via `IRoles.ListOwnerIds`, which masks anonymous owners from non-owner callers. Harmless before — `EditMembers` was effectively Owner-only, and an Owner caller sees no masking — but `Moderate` now implies `EditMembers`, so a Moderator saw an empty owner list for anonymous owners and sailed past the guard.

Fixed to use the unmasked backend path, the same one the message-moderation guards already use. This is exactly the failure mode the PR description called out for message deletion; I'd missed that granting `EditMembers` opened a second door onto it.

## Test coverage that was missing

Three capabilities had zero coverage despite being explicit requirements:

- `ModeratorCanKickMemberButNotOwner` + `ModeratorCannotKickAnonymousOwner` (regression test for the above)
- `ModeratorCanRestoreRemovedMessages`
- `PlaceModeratorModeratesPublicPlaceChats` — one appointment on the place root chat reaches every public place chat, and no private one

The place test failed on first run, but the feature was fine — my test built a standalone group chat instead of a place chat (`CreateChat` defaults `Kind = ChatKind.Group`; place chats need `Kind = null`). Worth knowing if you write similar tests.

| Suite | Result |
|---|---|
| `ActualChat.CI.slnf` build | clean |
| `Chat.IntegrationTests` (full) | **363 passed**, 0 failed |
| `ModeratorRoleTest` | 14 passed |
| `CallModerationTest` | 11 passed |

PR #4153, opened 2026-08-11
PTT gesture fixes 

alexis-kochetov:


alexis-kochetov:
Addressed the review: the proximity/guard transition logs are demoted to Information (they are expected transitions; Debug would be filtered out of client logs, defeating their field-diagnostics purpose). The FaceDown-fire log intentionally stays at Warning: it fires at most once per stop gesture and is the only on-device forensic record for gesture behavior — client observability is otherwise limited (RPC strips stacks, Sentry is MAUI-only).

PR #4181, opened 2026-08-17
feat(commands): ApiCommand idempotency + backward-compatible command deserializer

crui3er:
## Summary

Adds **command idempotency** for API commands plus a **backward-compatible, version-gated command deserializer** so the two can roll out safely.

A new base record `ApiCommand` carries a client-generated `Uuid` (idempotency key, `Key 0`). A server-side CommandR filter (`ApiCommandDeduplicator`) claims each outermost `ApiCommand` by `(SessionHash, Uuid)` in an **in-process** store, runs it once, and replays the stored result to duplicates (client retries / reconnects).

**All 87 API commands are migrated** — every `IApiCommand` now derives from `ApiCommand<TResult>` as an all-`required init` record with its own members starting at `Key(2)`.

## What's included

- **`ApiCommand` base** (`Core`) — `required init` records, `Uuid` auto-generated so callers never pass it; swappable `NewUuidGenerator`. All 87 commands migrated (273 declaration edits, 612 call sites incl. `.razor`, 71 `var (a, b) = command;` deconstructions rewritten to property reads).
- **`IdempotencyStore`** (`Core.Server`) — a singleton `ConcurrentDictionary` of claims; an entry carries an expiry and a `TaskCompletionSource`, so a duplicate wakes the moment the owner finishes instead of polling. `InProgressTtl` 5 min, `CompletedTtl` **15 min**, plus a prune sweep and a `MaxEntryCount` cap that bound the memory.
- **`ApiCommandDeduplicator`** — the CommandR filter: claim → run / replay / await → complete or release. Failures aren't cached: the claim is released, which wakes waiters and frees the key.
- **`INotDeduplicated` opt-out** — commands whose repeat is harmless but whose suppression isn't, or that are frequent enough to crowd the store: `UserPresences_CheckIn`, `ChatPositions_Set`, `Uploads_Append`, `ChatUsages_RegisterUsage`, `UserSettings_Set`, `Notifications_RegisterDevice`.
- **Version-gated deserializer** (`ApiCommandRpcArgumentSerializer`, `Core`) — decorates every **server** inbound msgpack arg serializer. Pre-`UuidVersion` (**2.17**) clients send the legacy layout (Session @ 0); the decorator prepends an empty `Uuid`, gating on the API version read per-peer from the RPC handshake's `RemoteApiVersionSet`. Backend peers (BackendScope, no ApiScope), map payloads (own `[MessagePackFormatter]`, and the keyless formats TS clients use) and non-command args pass through verbatim.
- **Client-side `Uuid` reuse** — `SendingMessages` now resends the queued request's `Uuid` instead of building a fresh command, so a retry after its 15 s timeout replays the stored result rather than posting a second message (the duplicate-attachment bug).
- **Observability** — `IdempotencyMeters` on `CoreServerInstruments.Meter`: `command.dedup.{outcome,overrun,release,result_size}`.
- **`ApiCommand` moved `Api` → `Core`** so the decorator can gate on `typeof(ApiCommand).IsAssignableFrom(type)` (no Api-only deps; namespace unchanged).
- **Docs** — `docs/architecture/command-idempotency.md` (design, TTLs, opt-outs, the legacy path) + updates to `overview.md` / `service-design.md`.

## Key decisions

- **~95% is enough** — an occasional double-run during network trouble is acceptable ("we're not transferring money"); dedup state is not stored atomically with the operation's DB.
- **In-process, not distributed.** The store originally ran on Redis with mesh-liveness reclaim of a dead owner. That was dropped: a client's commands ride one RPC connection to one node, so the cross-node duplicate the distributed store existed for is the rare case — not worth a runtime dependency, two network round-trips per command, and command results living in shared infrastructure. A node restart or a reconnect elsewhere loses the window and simply re-runs the command, exactly as before dedup existed.
- Dedup key = **session hash + Uuid** (doesn't leak the session, prevents cross-session collision).
- Carry the `Uuid` **on the command** (base type), not via Fusion headers.
- Backward compat gated by the **API version from the RPC handshake** (per-peer), not a length/exception heuristic; `UuidVersion` pinned to the shipping release (2.17) — re-check it against `version.json` if this branch misses another release.
- **No positional constructors** on commands — `Uuid` must never appear at a call site, and an all-`init` record is the only shape that round-trips through MessagePack, Newtonsoft and System.Text.Json alike.

## Testing

- **Unit** — `IdempotencyStoreTest` (claim / replay / waiter wake-up / release / both TTLs / cap), `ApiCommandDeduplicatorTest` (dedup, opt-out, outcome metrics via `MeterListener`), decorator transform (legacy / current / null-version / map payloads / keyless), command serialization round-trips.
- **Integration** — `ReactionDeduplicationTest`, `MessageResendDeduplicationTest` (same `Uuid` resend posts one message; a fresh `Uuid` posts a second), `ApiCommandVersionSeamTest` (the decorator resolves on every inbound server msgpack format).
- **Live manual verification** — a real pre-`UuidVersion` Windows client against a new server: reactions work, the server logs the migration, and the deduplicator skips the empty-`Uuid` legacy command (old clients degrade gracefully, they just don't get idempotency).
- Full `NoMaui` solution build plus the MAUI heads (Android/Windows) build clean.

## Notes for review

- All server inbound msgpack arg deserialization now routes through the decorator; peers ≥ `UuidVersion` and non-command args delegate to the inner `RpcByteArgumentSerializerV4` verbatim (identical behavior).
- Old clients keep working during the rollout: their legacy-layout commands get an empty `Uuid`, which the deduplicator's `Uuid.Length > 0` guard skips.


alexyakunin:
@crui3er LGTM - I mostly read design doc & looked up at how command looks.

alexis-kochetov:
### Code review

Found 1 issue:

1. The `UuidVersion = new(2, 17)` legacy-client gate cannot distinguish builds within one minor-version cycle: `RpcDefaults.ApiVersion` comes from `ApiConstants.Version`, which is pinned to `Major.Minor` only (`X.Y.0.0`), and `version.json` was already `2.17-alpha` before this PR. Any client built from `dev` in that window (cached WASM tabs, mobile alpha builds) sends the *old* layout but reports `2.17.0.0` in the handshake, passes the `peerVersion >= UuidVersion` check, and skips the migration — the server then deserializes `Session` into the `Uuid` slot and the first domain field into `Session`, corrupting every `ApiCommand` from those clients until they update. Fix: bump `version.json` to a new minor version as part of landing this change (or gate on a version source with per-build granularity).

https://github.com/Actual-Chat/actual-chat/blob/0126b46fbdda2ee9e5f555b8ea7d576e19f4b233/src/dotnet/Core/Serialization/ApiCommandRpcArgumentSerializer.cs#L20-L37

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #4201, opened 2026-08-19
llive audio trim once

alexis-kochetov:


src/dotnet/UI.Blazor.App/Services/Audio/AudioStreamDemuxer.cs
```diff
@@ -88,10 +102,57 @@ protected override async Task OnRun(CancellationToken cancellationToken)
...
+                continue;
+
+            // Information, not debug: DebugMode.LiveStreaming is false in production, and a
+            // dropped track is the one thing here a listener can actually hear.
+            Log?.LogInformation(
+                "Dropping stream N{StreamIndex}: {BacklogMs}ms queued exceeds {LimitMs}ms",
+                index,
+                backlogFrames * Constants.Audio.FrameDurationMs,
+                Constants.Audio.MaxTrackBacklog.TotalMilliseconds);
+            while (entry.Channel.Reader.TryRead(out _)) { }
```
alexyakunin, line 131: It drops the whole track when its backlog is 2s? IMO a huge disadvantage, esp. if there is no video. You miss the whole phrase when you lag by 2s.

alexyakunin (changes requested):

PR #4202, opened 2026-08-19
feat(markup): Markdown-style tables + markup parser benchmarks

alexyakunin:
## Tables in the markup

A table is a **block markup**: a header row, a delimiter row that sets per-column alignment, and any number of body rows. Cells hold **inline markup only**, so a table can neither contain another block markup nor be nested into one (list item, block quote and header content is inline-only, so `| a | b |` inside them stays text).

```
| Build  | Platform | Startup, ms | Notes                        |
| ---    | :--      | --:         | :--                          |
| v2.9.2 | Android  | 1 615       | **-12%** after the mibc bump |
| v2.9.2 | iOS      | 2 210       | still the WASM boot          |
```

### Grammar (GFM, with one deviation)

- Delimiter cells: `---`, `:--`, `:-:`, `--:` (any number of dashes). The delimiter row must have the same cell count as the header row, otherwise it isn't a table and both lines stay a paragraph.
- Cell text is trimmed; a body row is padded or truncated to the header's cell count.
- `\|` is a literal pipe inside a cell; formatting escapes it back.
- The table ends at a blank line or at the first line that isn't a row.
- **Deviation:** a row must start with `|`. Markdown makes the leading pipe optional, but `|` is also the spoiler token here, so a pipe-less `a | b` line has to stay ordinary text. A line that merely starts with `|` (e.g. `||spoiler||`) becomes a table only when a matching delimiter row follows it.

### Model

`TableMarkup` (header + `TableColumnAlignment[]` + rows) → `TableRowMarkup` → `TableCellMarkup`, rectangular by construction: the constructor rejects a row whose cell count differs from the header's, and a cell whose content is block markup.

Wired through the visitor bases, both rewriters, the trimmer (drops whole rows — a partial row would break the shape), the validator, both extractors, and the readable / HTML / editor formatters. Flattened text (notifications, chat-list and quote previews) leaves the delimiter row out, since it's syntax rather than content.

Rendered by `TableMarkupView`; the table scrolls horizontally on its own so a wide one never widens the message. AOT helper sources were regenerated for the new serializable types.

**Tests:** 30 parser/format cases (alignment, padding, truncation, escapes, block ordering, the not-a-table cases, list/quote nesting) plus a MessagePack round-trip.

## Benchmarks

`tests/Core.Benchmarks` → `tests/Benchmarks`, now referencing `UI.App` instead of `Core`, so it can benchmark essentially any client-side code.

`MarkupParserBenchmarks` runs `MarkupParser` over three real Voxt messages and adds a **Chars/s** column. Current numbers (this machine, ShortRun, in-process):

| Sample  | Chars  | Mean     | Chars/s | Allocated |
| :--     | --:    | --:      | --:     | --:       |
| Regular | 338    | 335 µs   | ~1.0 M  | 78 KB     |
| Table   | 776    | 752 µs   | ~1.0 M  | 120 KB    |
| Long    | 12 872 | 5 213 µs | ~2.5 M  | 1.5 MB    |

Two things stand out: a short message costs ~0.3 ms and ~78 KB, and `Simplify()` is a small part of it (`ParseRawOnly` is within a few percent of `Parse` except on the long sample). A second parser implementation belongs here as another `[Benchmark]` method, which makes it directly comparable against the current one.

Every benchmark now runs **in-process** — BenchmarkDotNet 0.15.x doesn't know the .NET 11 runtime moniker, so its default csproj toolchain threw in SDK validation before a single benchmark could run.

## Note

`.claude/style-bypasses.md` gained one entry: BDN's in-process toolchain refuses to run a `sealed` benchmark class, so `MarkupParserBenchmarks` can't be sealed. Flagging it because bypasses are meant to be your call.

## Verification

- `dotnet build ActualChat.CI.slnf` ✓
- `Chat.UnitTests` 559 passed, `Chat.UI.Blazor.UnitTests` 340 passed
- `npm run build:Debug` ✓

https://claude.ai/code/session_01CGrcG5J94aXBRtEKwdxwu4


alexyakunin:
Superseded by #4203 — same commits, branch renamed to feat/table-markup.

PR #4204, opened 2026-08-19
test(benchmarks): port the markup parser to ParsecSharp and measure it

alexyakunin:
Answers "would ParsecSharp be faster than Pidgin for our markup grammar?" — **no**, and with numbers.

## What's here

**`ParsecMarkupParser`** — the whole grammar ported onto ParsecSharp, written as a literal drop-in: same namespace (`ActualChat.Chat`), same options (`UseUnparsedTextMarkup` / `MustSimplify` / `AllowIncompleteMarkup`), same `Markup` tree, same file layout as `MarkupParser.cs` so the two diff cleanly. Adopting it would be `git mv` + moving the `ParsecSharp` PackageReference from `Benchmarks.csproj` to `Api.csproj` + renaming the type.

**`MarkupParserEquivalence`** — because a speed number for a parser that doesn't produce the same tree is worthless. 71 corpus inputs covering every rule (urls, emails, mentions, hashtags, styles, incomplete styles, preformatted, code blocks, headers, lists, quotes, tables, blank-line runs, stray specials) × the 4 option combinations, compared as **structural dumps**, not formatted text — so a difference in tree shape can't hide behind an identical round-trip. The benchmark's `GlobalSetup` runs it on its own sample and throws on a mismatch; `dotnet run -c Release --project tests/Benchmarks -- verify` runs it alone.

It passes: `ParsecMarkupParser matches MarkupParser on all 71 corpus inputs.`

## The measurement

| Sample | Parser | Mean | Chars/s | Allocated |
| :-- | :-- | --: | --: | --: |
| Regular (338 ch) | Pidgin | 299 µs | 1.13 M | 78 KB |
| Regular | ParsecSharp | 822 µs | 0.41 M | 3.4 MB |
| Table (776 ch) | Pidgin | 701 µs | 1.11 M | 120 KB |
| Table | ParsecSharp | 1 454 µs | 0.53 M | 7.1 MB |
| Long (12 872 ch) | Pidgin | 4 302 µs | 2.99 M | 1.5 MB |
| Long | ParsecSharp | 10 437 µs | 1.23 M | 45.8 MB |

**2.4–2.8× slower, 30–59× more allocation.**

## Why

`ParsecSharp.Parser.Many` accumulates via `Enumerable.Append`:

```csharp
private static IParser<TToken, IReadOnlyCollection<T>> ManyRec<TToken, T>(
    IParser<TToken, T> parser, IReadOnlyCollection<T> result)
    => parser.Next(x => ManyRec(parser, result.Append(x)), result);
```

Every matched token allocates a LINQ append iterator plus a `CountableEnumerable` wrapper — ~2 allocations per character — and materializing the run walks a chain of *n* nested iterators. `TakeWhile(pred)` is literally `Many(Satisfy(pred))`, so the library offers no cheaper path for a character run. Pidgin's `ManyString` appends into a pooled `StringBuilder` instead.

To keep the trial fair I gave ParsecSharp the same optimization — `TakeWhileStringParser`, a custom `PrimitiveParser<char, string>` that walks the parse state directly — and used it for every char-run-to-string in the grammar. That's what the numbers above measure. Before it, the same port ran **3.0–3.6× slower with 42–70× the allocation** (long sample: 13.3 ms / 63.9 MB), so the primitive buys back roughly a fifth of the time and a third of the allocation; the remainder is spread across the other `Many` call sites (block sequences, inline element lists, ids, quoted names) and the continuation-passing core.

## What I'd take from it

- ParsecSharp is out for this grammar.
- The interesting number isn't the ratio, it's the absolute: **~300 µs and 78 KB to parse a 338-char message**. That's the combinator-library tax, not the grammar — `Simplify()` is a rounding error (`PidginRaw` is within a few percent of `Pidgin`).
- If a "much faster parser impl" is still wanted, the evidence points at a hand-written recursive-descent/scanner pass over the string, not at another combinator library. The equivalence harness in this PR is reusable as-is for that: point it at the new type and it will tell you whether it's honest before you look at its speed.

Nothing in `src/` changes — this is entirely under `tests/Benchmarks`, plus the `ParsecSharp` entry in `Directory.Packages.props`.

## Verification

- `dotnet build ActualChat.CI.slnf` ✓
- `dotnet run -c Release --project tests/Benchmarks -- verify` ✓
- `dotnet run -c Release --project tests/Benchmarks -- MarkupParserBenchmarks` ✓ (numbers above)

https://claude.ai/code/session_01CGrcG5J94aXBRtEKwdxwu4

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -196,6 +196,7 @@
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.2" />
     <PackageVersion Include="OpusSharp.Core" Version="1.6.1.3" />
     <PackageVersion Include="OpusSharp.Natives" Version="1.6.1.3" />
+    <PackageVersion Include="ParsecSharp" Version="5.2.0" />
     <PackageVersion Include="RestEase" Version="1.6.4" />
     <PackageVersion Include="Pidgin" Version="3.5.1" />
     <PackageVersion Include="Polly" Version="8.5.2" />
--- a/tests/Benchmarks/Benchmarks.csproj
+++ b/tests/Benchmarks/Benchmarks.csproj
@@ -33,6 +33,7 @@
 
   <ItemGroup>
     <PackageReference Include="BenchmarkDotNet" />
+    <PackageReference Include="ParsecSharp" />
   </ItemGroup>
 
 </Project>
--- a/tests/Benchmarks/MarkupParserBenchmarks.cs
+++ b/tests/Benchmarks/MarkupParserBenchmarks.cs
@@ -7,18 +7,23 @@ using BenchmarkDotNet.Running;
 
 namespace ActualChat.Benchmarks;
 
+// Run: dotnet run -c Release --project tests/Benchmarks -- MarkupParserBenchmarks
+// The equivalence check alone: dotnet run -c Release --project tests/Benchmarks -- verify
+// Setup fails the run when the parsers disagree on the sample, so a timing is never reported
+// for a parser that isn't equivalent.
+
 /// <summary>
-/// <see cref="MarkupParser"/> throughput on two real Voxt messages - see <see cref="MarkupSamples"/>.
-/// The "Chars/s" column is the number the parser is judged by; a second parser implementation
-/// should show up here as another [Benchmark] method rather than another run.
-/// Run: dotnet run -c Release --project tests/Benchmarks -- MarkupParserBenchmarks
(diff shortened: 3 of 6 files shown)


alexyakunin:
Closing: ParsecSharp was measured and rejected (2.4-2.8x slower, 30-59x more allocation - details in the description above, kept as the record). The follow-up is #4205, which makes the existing Pidgin parser ~2x faster instead.

PR #4220, opened 2026-08-21
fix: The main-thread stall cascade and the recovery paths it exposed

alexyakunin:
Started from one console log: a 7.25s main-thread stall on dev that took video
recording and playback down with it. A five-way investigation traced the chain,
and each defect found along the way is one commit. 39 commits, 45 files.

Every mechanism below was read from source; where something was measured or
verified in a browser, it says so.

---

## 1. The render storm — the root cause

**Streaming message render rate.**
*Problem:* `ChatEntryMessageInternalView` inherited the default `IUpdateDelayer`,
which collapses to `MinDelay` (~32ms) whenever `UIActionTracker` reports instant
updates — true for any UI action and for 300ms after one, so effectively always
during chat use. Each recompute becomes its own render batch, SignalR message and
synchronous DOM patch: ~30/sec per streaming message.
*Solution:* `FixedDelayer.Get(0.2)`, which ignores the action tracker and pins the
state to 5 updates/sec — the same 200ms grid the streaming markup already animates on.

**`image-skeleton` fought Blazor for the `class` attribute.**
*Problem:* Two writers owned `class`. Blazor rewrote it on every re-render from a
tree that only knew the initial state; the component restored its own state class in
response. Worse, `thumbnailLoaded()` claimed state `'thumbnail'` even on the branch
where it deliberately left `show-image-original` alone — so once a cached original
beat a proxied thumbnail, the next class write **permanently downgraded a sharp
image to its blurry thumbnail**, and `imageLoaded()` never fires again to undo it.
*Solution:* two commits. First the state desync, so the restore lands on the right
class. Then the conflict itself: state moved to `data-image-state`, which Blazor
never renders, seeded once from Razor through a constant `initial-state` property.
CSS moved to attribute selectors (same specificity), and `animation-sync` gained a
non-class matcher so the skeleton shimmer stays phase-aligned.
*Verified in-browser:* two consecutive Blazor-style class writes leave the state
untouched, the skeleton selector still resolves to `animation-name: pulse`, and a
class write costs **1 mutation record instead of 3**.

`connectedCallback()` had a third bug on the same field: it re-seeded `_imageState`
from the Razor-supplied `initial-state` on **every** attach, so a Blazor move of an
already-loaded image rolled it back to `skeleton` and re-showed the shimmer over a
decoded bitmap. It now seeds only while the state is `none`.

The rule is now written down in `docs/development/ui-components.md`: a Lit
component may *have* a `class` but must never *write* one.

---

## 2. VideoFrame lifetime

**Ceiling frame on downscaler rejection.** *Problem:* the `try/finally` around the
race cleared the hang timer but had no `catch`, and `downscaler.process()` closes
nothing on failure — so every rejection stranded a full-resolution frame. Rejection
is exactly what a lost GPU context produces, **at capture rate**. *Solution:* close
it on the rejection path; also close `produceCeiling`'s input from a `finally`, and
stop a throwing `preview.forward()` — a cosmetic tap — from stranding the ceiling
and every tier.

**Pulled bundle on drain throw.** *Problem:* both drain-and-yield points hold a
bundle the operator owns but has not handed to the encoders; the operator's
`finally` disposes encoders, not the bundle. 1-4 frames per encoder error.
*Solution:* close the bundle's layers before the error propagates.

**RPC transferables.** *Problem:* an `RpcCall`'s transferables arrive owned by the
receiving realm, and three paths dropped them: no call id, no such method, and — the
one that produced `A VideoFrame was garbage collected without being closed` in the
log — a disposed server, which restored a null `onmessage` and let in-flight
messages fall on the floor. *Solution:* release them on the non-delivery paths, and
stand in with a drain handler when there is no previous handler to restore. The
video recorder also defers `Worker.terminate()` by one turn so queued preview frames
reach that handler instead of dying with the worker.

**Keep-alive clone** and **`recorder-worker.pushFrame`.** *Problem:* both drop a
frame when the consumer or handler is gone — a bare `yield` with no guard, and an
optional-chained forward that silently discards when `deps.pushFrame` is absent.
*Solution:* guard both.

**WebGPU device loss.** *Problem:* the deferred-cleanup queue mixed GPU `destroy()`
closures (which must not run after device loss) with `VideoFrame.close()` closures
(which must). `onBlurDeviceLost` cleared the lot. *Solution:* split the queues —
drop the GPU one, drain the frame one.

---

## 3. Sender: encoder recovery

**The fatal-error callback only logged.** *Problem:* per spec a fatal WebCodecs error
closes the encoder and clears its queue without settling anything, so up to five
bundles x two layers stayed pending and nothing noticed until the 3s bundle watchdog.
*Solution:* fail the in-flight items from the callback.

**A closed encoder was a one-way door.** *Problem:* `resetCodec()` returns early on
state `'closed'` **silently**, so the watchdog's "resetting encoders in place"
reported a recovery it had not performed; and `getNotReadyError()` built a plain
`Error`, which falls outside `awaitPending()`'s all-reset sweep and is classified
fatal. The next `submitBundle()` after any fatal codec error was therefore
guaranteed to kill the pipeline. *Solution:* make the not-ready error recoverable,
have `resetCodec` say when it cannot reset, and rebuild any slot whose state is
`closed`/`unconfigured` before submitting, forcing a keyframe so the new encoder
anchors.

**The recovery sweep had no exit.** *Problem:* on that all-reset path every promise
settles, so the 3s bundle watchdog never fires. A codec failing on **every** encode
therefore skipped bundles forever while the pipeline looked healthy and produced
nothing. *Solution:* escalate to `ENCODER_INIT_FAILED` after 30 consecutive skips
with no encoded output; any fulfilled output resets the counter.

**A degraded adapter never recovered.** *Problem:* `degradeAndReset` dropped
`maxInflight` to 1 and nothing raised it again, so one timeout capped throughput for
the session — and with the operator still pipelining five deep, that permanently
armed a `queue full` rejection which was *also* a plain `Error`, i.e. fatal.
*Solution:* make queue-full recoverable and restore the original depth after 30 clean
outputs.

**…and then oscillated.** *Problem:* a machine marginally over budget re-degrades
right after every restore, so a fixed bar produces degrade → restore → timeout
roughly once a second, each turn costing a codec reset and a forced keyframe.
*Solution:* double the bar on each restore (capped at 3840, ~2 min at 30fps) — one
transient hiccup still recovers in a second, a persistently marginal encoder parks
at `maxInflight=1`.

**Reshape mixed two ladder generations.** *Problem (pre-existing):* the reshape sized
the encoder set from the **bundle's** layer count but read its configs from a fresh
controller snapshot. A ladder change landing between the bundle leaving downscale and
the reshape drain completing made those disagree — `cur.configs[i]` is not guaranteed
to exist for every `i < layerCount`, so the slot loop either built encoders from the
wrong generation's dims or threw on an undefined config and killed the pipeline.
*Solution:* `CapturedBundle` carries the `LadderState` version it was built from, and
a bundle the controller has moved past is dropped rather than reshaped.

**Recovery re-entrancy.** *Problem:* `recoveryScheduled` was cleared when the timer
fired, not when recovery finished. An error arriving during `recoverNow` passed the
guard; the overlapping `startWorker` rejected with "already running", which scheduled
another — a self-feeding loop to the attempt cap, which hands the session to
`engageEncoderFallback` and **permanently drops it to software H.264 at 640x360 or
lower**. *Solution:* hold the guard across the whole of `recoverNow` — and bound that
pass at 15s, since its awaits reach a worker that may itself be wedged and an
unbounded one would have retired recovery for the session.

---

## 4. Receiver: decoder and present recovery

**Teardown deadlock.** *Problem:* `pacedEncodedBuffer`'s consumer awaited the
producer unconditionally, and the producer only exits on upstream end or abort. A
wedge restart returns the consumer without raising either while the RpcStream keeps
delivering, so teardown waited on a loop with no reason to end. `[CODEC_EXHAUSTED]`
never escaped: `whenDone()` never rejected, `onError` never fired, and the 30s stall
timer was defeated because arriving chunks kept resetting it — leaving a main-thread
poller as the only thing that could break the wedge, at 6s of black video.
*Solution:* give the producer its own stop signal, raised by the consumer's `finally`.

**Recovery budget.** An earlier commit here reset `consecutiveRecoveries` on *any*
decoder output rather than only a paired one, to stop a merely slow decoder from
exhausting its budget. It was **dropped**: `8b32b6ac6f` on `dev` fixes that failure
from the submit side (the hang deadline no longer expires while waiting for a chunk
the decoder could accept), which is the better place for it, and the looser reset
would have let a decoder emitting stale frames never exhaust at all.

**MSTG present had no ceiling.** *Problem:* `writer.ready` is the pacer's only
backpressure signal and was awaited unbounded, so a wedged `<video>`/compositor
stalled presentation and the decode behind it indefinitely — a second wedge mode.
*Solution:* cap one wait at a second and drop the frame; `presentPacer` already
treats a false result as a presenter drop and closes the frame.

**Stat was stale exactly when quoted.** *Problem:* `encodedQueueCount` was written
only by the latency tap, which rides the decoded-frame stream at ~1Hz — so it froze
the instant decoding stopped, which is when a wedge report quotes it. The `buf=4` in
the incident log was 11.5s stale and reads as something that did not happen.
*Solution:* sample the buffer in `getStats`.

**Tests.** `paced-encoded-buffer.test.ts` is new and verified to fail against the
pre-fix code (the deadlock, with an "ends normally" control). Four more cover paths
the existing suites did not reach: the decoder's `CODEC_EXHAUSTED` budget and its
unpaired-frame close, the stale-ladder-generation drop, and the degrade-restore
backoff. Full suite: **540 passing**.

---

## 5. RPC, build, diagnostics, assets

**RPC first-handshake sweep.** *Problem:* the TS port collapsed .NET's three-state
`RpcPeerChangeKind` to a boolean and lost the `ChangedToVeryFirst` guard, so even a
cold connect ran reconnect processing and swept the very call it had just put on the
wire — the client reported failure for work the server had already executed.
*Solution:* vendored `actuallab-rpc` synced from Fusion `3f794818`, plus **Fusion
14.3.29 to 14.3.34** for the C# half.

**Stream-control calls had no timeout at all.** *Problem:* the TS service defs never
carried `[RpcMethod(ConnectTimeout = 10)]`, and `rpc-peer` skips the timeout sweep
entirely for a call with no `timeouts` — so `RequestKeyFrame`,
`ChangeRecordingQuality` and `ChangePlaybackQuality` could hang forever.
*Solution:* port the 10s connect timeout. Run stays unbounded, which is what .NET
resolves these query-kind methods to as well.

**A failed keyframe request burned the cooldown.** *Problem:* `requestKeyFrame`
stamped `lastKeyFrameRequestTime` before the call and swallowed the rejection, so one
failure blocked retries for the full 10s. *Solution:* stamp on success only, guarded
by a single expiring timestamp — an in-flight flag alone would have latched the hook
off for the life of the player, since the run timeout is unbounded.

**WebGL context loss was invisible — then over-reported.** *Problem:* no surface
listened for `webglcontextlost`, and `WebGlDownscaler` renders black **without
throwing** — so the `catch` that engages its Canvas2D fallback never ran, and a real
context loss was indistinguishable from an ordinary teardown in the logs.
*Solution:* check `isContextLost()` up front and register listeners on all three
surfaces — then **detach them in `dispose()`**, because `dispose()` ends with
`WEBGL_lose_context.loseContext()`, which dispatches the very event just registered.
Left attached, every ordinary teardown ran the loss handler: on `WebGlDownscaler`
that permanently disabled WebGL downscale for the session, and on the two blur
renderers it logged a phantom loss that masked the real ones.

**Hidden-tab throttling reported as a stall.** *Problem:* the watchdog only skipped
hidden ticks under a ~1s throttling profile, but Chrome throttles a background tab
into minute-long buckets and can freeze it outright — producing entries like
`recovered after 892825ms total stall`, which bury the real ones. *Solution:* skip the
check entirely while hidden; a delta measured there carries no information.

**Font preloads never matched the CSS.** *Problem:* preload hrefs went through
`@Assets`, which fingerprints them, while the bundled `@font-face` keeps the plain
path — so every preloaded font was downloaded **twice** and the first copy discarded.
*Solution:* match the URLs. *Measured in-browser before and after:* **~238 KB of
duplicate woff2 per cold load, now zero.** MAUI had the same defect twice —
`icon.woff2` pinned to a stale generated `?t=`, and forkawesome missing the CSS's
`?v=1.2.0`.

---

## Deliberately not done

- **Translation throttling** — written up as
  [`docs/plans/better-translation.md`](docs/plans/better-translation.md) instead. The
  whole-message stream publishes one wire item per LLM chunk with no rate limit; the
  realtime stream is throttled at 500ms, a period chosen for latency though its real
  cost is up to four LLM calls/sec. The plan records the concrete fix, three claims
  from the first analysis that turned out wrong, and a dead overload found on the way.
- **Deferred `Worker.terminate()`** — *left as a one-turn defer.* Acking the drain
  would be deterministic for a live worker, but this is the cosmetic preview tap, not
  a correctness path, and a wedged worker would not drain in any bounded window
  either.
- **`WRITER_READY_TIMEOUT_MS = 1000`** — *left fixed.* Deriving it from frame duration
  would make the tolerance a constant backlog depth, but the failure it guards is
  binary (stuck compositor vs. not), and 1s already sits well above any real hiccup
  and well below the 6s wedge restart.
- **`MaxBufferedUnacknowledgedRenderBatches = 100`** — *profiled, left alone.* A local
  A/B against 30 showed no regression, but that is a null result: on localhost the
  client ACKs inside the burst so the buffer never fills. The latency-independent
  bound settles it — a chat switch sends 150-195 SignalR frames, peaking at **151
  within a 100ms window**. The existing comment is justified.
- **`@key` on the `change-item` loop** — *profiled, not worth it.* 29.3s of live
  transcription: 864 mutation records, peak 388/sec. The change-item machinery is
  ~8.5% of that; the dominant contributor is VirtualList item churn, which `@key` does
  not touch. An earlier claim that this could take the storm to zero does not hold.

## Verification

`tsc --noEmit`, `eslint` and the 540-test unit suite green; `dotnet build` succeeds
for `App.Server` and `UI.Blazor.App`. The UI-visible changes were exercised in a real
browser against the running dev server. The video failure paths were not — they need
a GPU stall to reproduce.

The branch is rebased onto current `origin/dev`, so it already sits on top of
`8b32b6ac6f` and `022d0f6459`.

### DIFF
--- a/src/dotnet/UI.Blazor.App/Components/ChatView/Items/ChatEntryMessageInternalView/ChatEntryMessageInternalView.razor
+++ b/src/dotnet/UI.Blazor.App/Components/ChatView/Items/ChatEntryMessageInternalView/ChatEntryMessageInternalView.razor
@@ -206,6 +206,10 @@
             IsUnreadByOthers, IsUnreadByOthersBlockStart);
         return new() {
             InitialValue = initialData, // To show the original text immediately
+            // The default delayer collapses to ~32ms whenever UIActionTracker reports
+            // instant updates — ~30 render batches/sec per streaming message, each one a
+            // synchronous DOM patch. 0.2s is the grid the streaming markup already animates on.
+            UpdateDelayer = FixedDelayer.Get(0.2),
             Category = GetStateCategory(GetType()),
         };
     }
--- a/src/dotnet/UI.Blazor.App/Components/VideoPanel/video-player.ts
+++ b/src/dotnet/UI.Blazor.App/Components/VideoPanel/video-player.ts
@@ -250,7 +250,8 @@ export class VideoPlayer {
 
     // PLI: receiver-requested keyframe (kept as a courtesy hook — most of
     // this is now driven by the worker's epoch-reset operator).
-    private lastKeyFrameRequestTime = 0;
+    // Parking expires after the cooldown — an unsettled call (run timeout is unbounded) must not latch it.
+    private keyFrameRequestBlockedSince: number | null = null;
     private readonly keyFrameRequestCooldownMs = 10000;
 
     // Render-quality hint state.
@@ -952,10 +953,10 @@ export class VideoPlayer {
     }
 
     private requestKeyFrame(): void {
-        const now = performance.now();
-        if (now - this.lastKeyFrameRequestTime < this.keyFrameRequestCooldownMs)
+        const sentAt = performance.now();
+        if (this.keyFrameRequestBlockedSince !== null
+            && sentAt - this.keyFrameRequestBlockedSince < this.keyFrameRequestCooldownMs)
             return;
-        this.lastKeyFrameRequestTime = now;
 
         infoLog?.log(`requestKeyFrame: stream=${this.streamId}`);
         // Worker-side hint (the new contract carries this, but the host
@@ -964,8 +965,22 @@ export class VideoPlayer {
         if (this.playerWorker)
             void this.playerWorker.requestKeyframe(this.streamId)
                 .catch(() => { /* worker stub is a no-op; ignore */ });
-        streamingApi.liveVideoStreams.RequestKeyFrame(RPC_SESSION_DEFAULT, this.streamId)
-            .catch((e: unknown) => warnLog?.log('RequestKeyFrame error:', e));
+
+        this.keyFrameRequestBlockedSince = sentAt;
+        // `sentAt` doubles as this request's identity, so a straggler settling after
+        // its parking expired can't disturb whatever request replaced it.
+        void streamingApi.liveVideoStreams.RequestKeyFrame(RPC_SESSION_DEFAULT, this.streamId)
+            .then(
+                () => {
+                    if (this.keyFrameRequestBlockedSince === sentAt)
+                        this.keyFrameRequestBlockedSince = performance.now();
+                },
+                (e: unknown) => {
+                    // Only a request the server actually got starts the cooldown.
+                    warnLog?.log('RequestKeyFrame error:', e);
+                    if (this.keyFrameRequestBlockedSince === sentAt)
+                        this.keyFrameRequestBlockedSince = null;
+                });
     }
 
     public async startPull(streamId: string, skipToMs: number): Promise<void> {
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,7 +4,7 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
   </PropertyGroup>
   <PropertyGroup>
-    <ActualLabFusionVersion>14.3.29</ActualLabFusionVersion>
+    <ActualLabFusionVersion>14.3.34</ActualLabFusionVersion>
     <!-- .NET 11 is still in preview: every Microsoft.* package below shares this build number -->
     <DotNetVersion>11.0.0-preview.7.26381.103</DotNetVersion>
     <AspNetCoreVersion>$(DotNetVersion)</AspNetCoreVersion>
--- a/docs/development/ui-components.md
+++ b/docs/development/ui-components.md
@@ -141,6 +141,57 @@ body.hoverable .amazing-panel > button:hover,
 }
 ```
 
+## Lit Components: Blazor owns `class`
+
+A Lit custom element rendered from Razor has **two** potential writers for its host
+attributes — Blazor, which rewrites the element's markup on every re-render, and the
+component itself. Anything both of them write ends up in a permanent tug-of-war,
+because Blazor's render tree never learns what the component changed.
+
+**The rule: a Lit component may *have* a `class`, but must never *write* one.**
+`class` belongs to Blazor. The component owns only attributes Blazor does not render.
+
+Publish component state through a `data-*` attribute and match it from CSS —
+attribute selectors have the same specificity as class selectors, so nothing in the
+cascade changes:
+
+**Wrong** — the component writes `class`, Blazor overwrites it on the next render,
+the component writes it again, forever:
+```typescript
+@property({ reflect: true }) class: string;   // makes every foreign class write re-render
+
+updated(changed: Map<string, unknown>) {
+    if (changed.has('class'))
+        this.classList.add(`show-image-${this._state}`);   // fights Blazor
+}
+```
+```css
+image-skeleton.show-image-original .image { ... }
+```
+
+**Correct** — Blazor owns `class`, the component owns `data-image-state`:
+```typescript
+private applyState(): void {
+    this.setAttribute('data-image-state', this._state);
+}
+```
+```css
+image-skeleton[data-image-state="original"] .image { ... }
+```
+
+Razor keeps writing `class` exactly as it would for any other element:
+```razor
+<image-skeleton class="pic-image @ExtraClass" src="@url" />
+```
+
+If the component needs an **initial** state from Razor, take it as a Lit property
+(`@property({ attribute: 'initial-state' })`) rather than reading it back out of the
+attribute the component itself writes. Razor renders that property as a constant, so
+re-renders write the same value and never conflict.
+
+The same reasoning applies to any attribute Razor emits — `style`, `title`, `id`. If
+Razor renders it, the component must not write it.
+
 ## No Inline Tailwind in Razor
 
 Do NOT write Tailwind utility classes directly in `.razor` markup. Instead, assign a CSS class and use `@apply` in the CSS file.
--- a/docs/live-video/02-sender.md
+++ b/docs/live-video/02-sender.md
@@ -323,7 +323,9 @@ closures, no track references.
 
 | Symptom | Where it's caught | Recovery |
 |---|---|---|
-| Encoder hangs on `output()` | `AsyncVideoEncoder` timeout (3 s first frame, 1 s steady) | Reset, force keyframe, retry bundle |
+| Encoder hangs on `output()` | `encode.ts` bundle watchdog (3 s per bundle; per-item adapter timeouts are disabled) | `handleEncoderHang()` — fail pending recoverably, `reset()` + reconfigure, force keyframe, skip the bundle. Throws after 2 consecutive hangs |
+| Encoder reports a fatal error | WebCodecs `error` callback in `recorder-worker-host.ts` | Fail the in-flight items at once, then `encode.ts` rebuilds any slot left `closed`/`unconfigured` and forces a keyframe |
+| Adapter degraded by a timeout or out-of-order output | `CodecToAsyncAdapter.degradeAndReset` | `maxInflight` drops to 1, restored after `DEGRADE_RECOVERY_OUTPUTS` clean outputs |
 | Frame dims ≠ encoder config | `dim-mismatch-guard.ts` | Drop frame; counter tick |
 | Downscaler hangs (`process()`) | `downscale.ts` 1.5 s watchdog | Close + recreate, force keyframe; bail after 4 consecutive |
 | HW NVENC slot lost | next `acquire()` on pool | Pool recreates encoder, runs through `handleEncoderReset` |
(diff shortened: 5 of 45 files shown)


src/dotnet/UI.Blazor.App/Components/ChatView/Items/ChatEntryMessageInternalView/ChatEntryMessageInternalView.razor
```diff
@@ -207,6 +207,10 @@
         return new() {
             InitialValue = initialData, // To show the original text immediately
             Category = GetStateCategory(GetType()),
```
alexyakunin: Category should be the last property assigned here.

src/dotnet/UI.Blazor.App/Components/VideoPanel/video-player.ts
```diff
@@ -964,8 +965,22 @@ export class VideoPlayer {
         if (this.playerWorker)
             void this.playerWorker.requestKeyframe(this.streamId)
                 .catch(() => { /* worker stub is a no-op; ignore */ });
-        streamingApi.liveVideoStreams.RequestKeyFrame(RPC_SESSION_DEFAULT, this.streamId)
-            .catch((e: unknown) => warnLog?.log('RequestKeyFrame error:', e));
+
+        this.keyFrameRequestBlockedSince = sentAt;
+        // `sentAt` doubles as this request's identity, so a straggler settling after
+        // its parking expired can't disturb whatever request replaced it.
+        void streamingApi.liveVideoStreams.RequestKeyFrame(RPC_SESSION_DEFAULT, this.streamId)
```
alexyakunin, line 972: Is this call has timeouts? It's important due to .then.

src/dotnet/UI.Blazor.App/Components/ChatView/Items/ChatEntryMessageInternalView/ChatEntryMessageInternalView.razor
alexyakunin: Fixed — `Category` is now the last property, after `UpdateDelayer`, matching `AudioDiagnosticsModal.razor` and the rest. Folded into the `perf(ui): pin streaming message renders to a fixed delayer` commit rather than added on top, so the branch stays one-commit-per-item.

src/dotnet/UI.Blazor.App/Components/VideoPanel/video-player.ts
alexyakunin: Partly — and the gap is exactly why this is a timestamp rather than an in-flight flag.

`RequestKeyFrame` now carries `StreamControlTimeouts = new RpcCallTimeouts(10_000)` in `streaming-api.ts`, and `RpcCallTimeouts(connectTimeoutMs, runTimeoutMs = Infinity)` — so it's **connect timeout 10s, run timeout unbounded**. That mirrors what .NET resolves `[RpcMethod(ConnectTimeout = 10)]` on a query-kind method to. In `rpc-peer._maintainOutboundCalls` the connect branch fires while `sentAt === 0`; the run branch is `now - sentAt >= Infinity`, i.e. never.

So there are three outcomes, not two:

1. **Server answers** → first handler; cooldown restarts from the response time.
2. **No connection within 10s** → call rejects → error handler → `keyFrameRequestBlockedSince = null`, so the next hook retries immediately. This is the case you spotted, and it's the one that mattered: before, the cooldown was stamped before the call and the rejection swallowed, so a single failure blocked retries for the full 10s.
3. **Sent, server never answers** → neither handler runs, ever.

Case 3 is why `keyFrameRequestBlockedSince = sentAt` is assigned *before* the call: the cooldown expires on its own after `keyFrameRequestCooldownMs` (10s) with no handler involvement. An `isKeyFrameRequestInFlight` boolean would have latched the hook off for the life of the player there.

The `=== sentAt` guard on both handlers covers the fourth case — a straggler from case 3 settling long after its cooldown expired and a newer request replaced it.

PR #4245, opened 2026-08-25
Android ANR fixes: FCM cold-start path (A1-A4 + instrumentation)

alexis-kochetov:
Implements steps A1–A4 plus the instrumentation from
[docs/plans/android-anr-issues.md](../blob/feat/android-anr-startup/docs/plans/android-anr-issues.md):
66% of our Android ANRs are an FCM push waking a dead process and
`Application.onCreate` missing the 10s foreground-broadcast budget on
A51-class devices.

- **A1** — the dismissal/badge push goes out at `Priority.Normal`: it's silent
  and non-urgent, yet it cold-started dead apps on the 10s clock and burned
  the high-priority allowance real notifications need.
- **A2** — `WarmUpWebView` moved from `MainApplication.OnCreate` (every
  process start) to `MainActivity.OnCreate`: on a push start no WebView is
  coming, and its eager ThreadPool/JIT/JNI work is the leading suspect for
  the runtime mutex the main thread blocks on in the top cluster's dump.
- **A3** — the armed-service start in `Application.OnCreate` is skipped on
  Android 12+ (40 of 67 CSV events), where a background FGS start is a
  guaranteed-failure binder round-trip; MainActivity and the PTT wake path
  (FCM exemption window) still raise it where it can succeed.
- **A4** — WorkManager's androidx.startup initializer is removed from the
  merged manifest (workdb SQLite + four `WM.task-*` threads on every process
  start, zero WorkManager use in our code). The package itself can't go —
  `Xamarin.Google.AI.Edge.LiteRT` pulls it transitively — so `MainApplication`
  implements `Configuration.IProvider` for lazy on-demand init instead.
  Verified in the merged manifest; ProfileInstaller/ProcessLifecycle/
  EmojiCompat initializers untouched.
- **Instrumentation** — a `start reason` (process importance) mark and an
  `Application.OnCreate completed` mark, and `MauiStartupBreadcrumbs` now
  buffers marks and flushes off the main thread (the old per-mark synchronous
  append fed the stalls it measured). One dev/beta cycle should show where
  the 10s actually goes before we commit to B1 (lazy idle-primed MauiApp).

Deliberate trade-off: on a dozing device the badge/dismissal update may now
arrive at the next maintenance window instead of immediately.

Crashlytics (Android) stays — confirmed deliberate. A5/B1 deferred until the
breadcrumb data is in.

https://claude.ai/code/session_012mqjvfB5bQztHYqiSW4bak

### DIFF
--- a/src/dotnet/App.Maui/MauiProgram.cs
+++ b/src/dotnet/App.Maui/MauiProgram.cs
@@ -42,6 +42,9 @@ public static partial class MauiProgram
     {
         using var _1 = Tracer.MethodRegion();
         MauiStartupBreadcrumbs.Add("CreateMauiApp");
+#if ANDROID
+        MauiStartupBreadcrumbs.Add($"start reason: {AndroidUtils.GetProcessStartReason()}");
+#endif
 
         // Parse -t <seconds> for auto-shutdown (used for AOT testing)
         var args = Environment.GetCommandLineArgs();
--- a/src/dotnet/App.Maui/Platforms/Android/MainApplication.cs
+++ b/src/dotnet/App.Maui/Platforms/Android/MainApplication.cs
@@ -22,6 +22,9 @@ public sealed class MainApplication : MauiApplication, AndroidX.Work.Configurati
     public override void OnCreate()
     {
         base.OnCreate();
+        // The moment the main looper is free to dispatch the broadcast that may have started us;
+        // the delta from "CreateMauiApp completed" is pure MAUI framework overhead.
+        MauiStartupBreadcrumbs.Add("Application.OnCreate completed");
         // Any process start, not only a user launch. A message push starts us with no MainActivity
         // and no Blazor scope, so nothing else raises the armed service - leaving no PTT badge and
         // no media session for the headset button. Android 12+ bans foreground-service starts from
--- a/src/dotnet/App.Maui/Platforms/Android/MainApplication.cs
+++ b/src/dotnet/App.Maui/Platforms/Android/MainApplication.cs
@@ -7,44 +7,31 @@ namespace ActualChat.App.Maui;
 #pragma warning disable // Can be static
 
 [Application]
-public class MainApplication : MauiApplication
+public sealed class MainApplication : MauiApplication, AndroidX.Work.Configuration.IProvider
 {
+    // On-demand WorkManager init: AndroidManifest.xml removes WorkManagerInitializer to keep its
+    // db + threads off the process-start path, and this keeps WorkManager.getInstance() working
+    // for the transitive dependencies that may still call it.
+    public AndroidX.Work.Configuration WorkManagerConfiguration
+        => new AndroidX.Work.Configuration.Builder().Build();
+
     public MainApplication(IntPtr handle, JniHandleOwnership ownership)
         : base(handle, ownership)
         => Android.Util.Log.Info(MauiDiagnostics.LogTag, "---- Started ----");
 
     public override void OnCreate()
     {
-        WarmUpWebView();
         base.OnCreate();
         // Any process start, not only a user launch. A message push starts us with no MainActivity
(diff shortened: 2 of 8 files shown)


src/dotnet/App.Maui/Platforms/Android/MainApplication.cs
```diff
@@ -7,44 +7,34 @@ namespace ActualChat.App.Maui;
 #pragma warning disable // Can be static
 
 [Application]
-public class MainApplication : MauiApplication
+public sealed class MainApplication : MauiApplication, AndroidX.Work.Configuration.IProvider
 {
+    // On-demand WorkManager init: AndroidManifest.xml removes WorkManagerInitializer to keep its
+    // db + threads off the process-start path, and this keeps WorkManager.getInstance() working
+    // for the transitive dependencies that may still call it.
+    public AndroidX.Work.Configuration WorkManagerConfiguration
```
alexyakunin, line 15: Why this piece goes before .ctor?

src/dotnet/App.Maui/MauiProgram.cs
```diff
@@ -42,6 +42,9 @@ public static MauiApp CreateMauiApp()
     {
         using var _1 = Tracer.MethodRegion();
         MauiStartupBreadcrumbs.Add("CreateMauiApp");
+#if ANDROID
+        MauiStartupBreadcrumbs.Add($"start reason: {AndroidUtils.GetProcessStartReason()}");
```
alexyakunin, line 46: Why it uses lower case? It breaks any logging-like pattern. And I checked other MauiStartupBreadcrumbs.Add , there are many other violations. Let's use normal sentences-like casing, i.e. "Start readon: ..." - everywhere in breadcrumbs.

src/dotnet/App.Maui/Platforms/Android/MainApplication.cs
alexyakunin: Fixed in 59bd6b44a0 (on `dev`, since this is already merged).

No good reason — I put it there following the "instance properties before the constructor" rule in `docs/CODING_STYLE.md`, but that rule is about state and DI-injected properties. `WorkManagerConfiguration` is neither: it is a computed `IProvider` implementation that builds a fresh `Configuration` on every call, so it is a method in property clothing and belongs with the methods.

It now sits after `OnCreate` and before the `protected` `CreateMauiApp`, which keeps public-before-protected. The comment stays above the declaration — it explains why the member exists at all, which is the declaration-annotating case in the style guide rather than the body case.

src/dotnet/App.Maui/MauiProgram.cs
alexyakunin: Fixed in 59bd6b44a0 (on `dev`, since this is already merged). You are right, and I swept every breadcrumb, not just this one:

| Before | After |
|---|---|
| `start reason: {…}` | `Start reason: {…}` |
| `process started` | `Process started` |
| `static services warmed up` | `Static services warmed up` |
| `app rendered` | `App rendered` |
| `foreground` | `Became foreground` |
| `background` | `Became background` |

The bare `foreground`/`background` marks got a verb as well — as one-word lowercase fragments they read as a state label rather than an event, which is the same problem in a different shape. They sit right next to `Log.Info(…, "OnBecameForeground")` calls, so the naming lines up now.

Left alone because they already start with a code identifier and read as proper sentences: `CreateMauiApp`, `CreateMauiApp completed`, `MauiApp configured`, `MauiApp built`, `MainActivity.OnCreate (isFirstTime: {…})`, `Application.OnCreate completed`. I kept `isFirstTime:` lowercase inside that last one — it is a named-argument-style reference to the field, not sentence prose.

Also updated the one stale `start reason:` quote in `docs/plans/android-anr-issues.md`. Verified with `dotnet build src/dotnet/App.Maui/App.Maui.csproj -f net11.0-android` — 0 errors.

PR #4252, opened 2026-08-26
feat: Tooltip severity + auto-show, and a single recording status

alexyakunin:
## Why

The record button carried three tooltips at once — a bespoke status bubble, a `data-tooltip` on the wrapper repeating the same status text, and another on the button itself. Pressing a control on a desktop also hovers it, so two of them could be on screen together. And every press announced "Starting recording..." then "Recording - English" whether or not anything was wrong, duplicating what the chat already shows below.

The goal: tooltips appear on their own **only when something is actually broken**, and when they do, they say something the user can act on.

## 1. `feat(ui): add tooltip severity and hover-free auto-show`

`TooltipHost` only reacted to `mouseover`, so anything needing to show a tooltip by itself grew its own markup beside the shared one, and touch devices had no path at all. Two new attributes on the shared host:

| Attribute | Meaning |
|---|---|
| `data-tooltip-severity` | `error` → pink bg + red frame (reuses the `banner-error` tokens) |
| `data-render-script-tooltip-auto-show` | Non-empty shows that text with no pointer involved; empty clears it |
| `data-tooltip-auto-show-duration` | Absent → 3s normal / 10s error; explicit `0` → until the value clears |

Auto-show rides the existing `data-render-script-*` mechanism, so no interop call and no `ElementReference` — and its dedupe is already right, since the value *is* the text. Hover wins for display; an auto-shown tooltip re-asserts itself when hover stops holding. A shown tooltip tracks its trigger's attributes, without which a control changing state under a stationary pointer keeps stale text and never repaints on escalation to error.

## 2. `feat(audio): report recording status from one compute method`

`ChatAudioUI.GetRecordingStatus` now decides what any recording control should say. An unnamed problem stays `Starting`/`Reconnecting` until it outlasts `RecordingProblemGracePeriod` — a healthy start passes through exactly those states, so without the grace every press would flash an error. The grace runs from the later of the press and the last recorder-pipeline transition; both are needed, since `StartRecording` runs bounded startup waits before the recorder moves at all, and a mid-session disconnect has no press behind it.

## 3. `feat(audio): name the microphone failure instead of guessing at it`

Every layer already knew why the mic didn't open and discarded it at the boundary. Notably **`RecorderStartResult.NoPermission` was unreachable on web** — `startRecording` returned `true` or threw, never `false` — so every web failure became "Failed to start the recording". Now each layer reports what it knows:

- **web** — `startRecording` returns `"<result>:<DOMException name>"` instead of throwing, since an exception reaches Blazor as an opaque `JSException`
- **Android** — min-buffer, `AudioRecord` state and ctor failures separately
- **Windows** — `AccessDenied` and `DeviceNotAvailable` off the WinRT statuses
- **Apple** — builds its graph inside the capture iterator, so it can only fail while enumerating; reported via a new `IAudioRecorderBackend.OnRecordingFailed`

`AudioRecorderState.LastFailure` carries the verdict up. It's cleared only once recording actually runs, so it survives the recorder's restart loop — clearing per attempt would blink the reason off and on for as long as it kept failing. A named failure is reported with no grace, and before the "is this chat recording" check, because a permission refusal never gets as far as recording.

Messages: **No microphone access** / **No microphone found** / **Another app uses microphone** / **Unknown error (NotReadableError)**, plus **Reconnecting...** for connectivity. Tooltips now wrap instead of truncating (an error ending in an ellipsis isn't actionable) and break long platform codes so they stay inside the bubble.

## Verification

- `tsc --noEmit`, `eslint`, `node build.mjs` clean
- `dotnet build ActualChat.CI.slnf` clean; 695 UI unit tests pass, incl. 16 localization
- MAUI built per target: **android** ✅, **windows** ✅, **maccatalyst** ✅ (on the Mac, `-t:Compile` — its Xcode is 26.0 and the .NET Apple SDK wants 26.6, unrelated to this branch)
- Tooltip appearance checked by rendering the compiled bundle CSS: severity colours, wrapping, 3-line clamp, and long-token breaking all verified

**Not yet exercised in the running app** — no `/server-loop` on this worktree. The four failure paths still want a live check.

https://claude.ai/code/session_01BUB6VAFPEF2SHE2ds3RN4i


src/dotnet/App.Maui/MaciOS/Audio/AppleAudioCapture.cs
```diff
@@ -42,8 +42,10 @@ public static bool IsInputNodeHeld {
     private AudioFocusUI AudioFocusUI => field ??= hub.AudioFocusUI;
     private ILogger Log => field ??= hub.Services.LogFor(GetType());
 
-    public Task<IAsyncEnumerable<IMemoryOwner<float>>?> Capture(CancellationToken cancellationToken)
-        => Task.FromResult<IAsyncEnumerable<IMemoryOwner<float>>?>(CaptureInternal(cancellationToken));
+    public Task<AudioCaptureResult> Capture(CancellationToken cancellationToken)
+        // Always reports success: AVAudioEngine setup happens inside the iterator, so a failure
```
alexyakunin: Really? See src/dotnet/App.Maui/Platforms/Android/Audio/AndroidAudioCapture.cs

alexyakunin: And src/dotnet/App.Maui/Platforms/Windows/Audio/WindowsAudioCapture.cs also doesn't always report success, right?

src/dotnet/App.Maui/Services/Recording/MauiRecorderEngine.cs
```diff
@@ -322,10 +325,21 @@ private void NotifyStateChange()
 
     private async Task ProcessAudioStream(IAsyncEnumerable<IMemoryOwner<float>> frames, CancellationToken cancellationToken)
     {
+        ChatId? chatId;
+        lock (_sync)
```
alexyakunin: We typically call such fields as _lock.

src/dotnet/UI.Blazor.App/Components/ChatAudioPanel/RecorderToggle.razor
```diff
@@ -9,17 +9,21 @@
...
-    // Steady recording (true, true, true) shows no tooltip - the red button + equalizer already signal
-    // recording, and the language is shown in the voice settings button right next to it. Only the
-    // transient/degraded states get one, and each persists until its condition clears.
-    var (showStatus, statusText) = (m.IsRecordingHere, m.IsMicrophoneCaptured, m.IsConnected) switch {
-        (false, _, _) => (false, ""),
-        (_, false, _) => (true, L.Call_StartingRecording),
-        (_, _, false) => (true, L.Reconnect_Title),
-        _ => (false, ""),
+    var status = m.Status;
+    var tooltip = status.Kind switch {
```
alexyakunin: Let's convert it to some method on status, maybe?

src/dotnet/App.Maui/MaciOS/Audio/AppleAudioCapture.cs
alexyakunin: You're right, the comment was misleading.

Android does its device setup **eagerly**: `GetMinBufferSize`, the `AudioRecord` ctor and the `State != Initialized` check all run before it returns `Ok(Enumerate(...))`, and each returns `Failed(...)` with its own code. Its `Enumerate` can then fail late as well — exactly like Apple's. So "fails while enumerating" isn't what distinguishes Apple; all three backends can do that.

What actually distinguishes it: Apple has **no eager path at all**. `CaptureInternal` does the entire AVAudioEngine setup — `SetVoiceProcessingEnabled`, `GetOutputFormat`, the engine start — inside the iterator, so `Capture()` has nothing it *could* report `Failed` for.

Reworded in 3e86c4339d:

```csharp
// No eager failure path, unlike Android and Windows, which open the device before returning
// their iterator: here the whole AVAudioEngine setup is inside CaptureInternal.
```

alexyakunin: Correct — Windows reports eagerly too. `AudioGraph.CreateAsync` (`DeviceNotAvailable` / `FormatNotSupported`), `CreateDeviceInputNodeAsync` (which is where the real `AccessDenied` comes from) and the `graph.Start()` block all return `AudioCaptureResult.Failed(...)` before the iterator is handed back — and its `Enumerate` can still fail late afterwards.

So Windows and Android behave the same way here, and the old comment implied Apple was alone in failing late. The wording fix in 3e86c4339d covers this one as well: it now contrasts Apple against *both* Android and Windows on the absence of an eager path, rather than on when the failure surfaces.

src/dotnet/App.Maui/Services/Recording/MauiRecorderEngine.cs
alexyakunin: Renamed in 3e86c4339d — 9 occurrences in `MauiRecorderEngine`.

Checked the convention before doing it: `_lock` 41, `_sync` 2 across `src/dotnet` — this file was one of the two outliers.

src/dotnet/UI.Blazor.App/Components/ChatAudioPanel/RecorderToggle.razor
alexyakunin: Done in 3e86c4339d — it's now `RecordingStatus.GetTooltip(IStringLocalizer l)`, and the component is one line:

```razor
var tooltip = status.GetTooltip(L);
```

It takes the localizer as a parameter rather than reaching for one, matching `ConversationParticipantsFormatter.GetText`. `FailureCode` is read off `this`, so the `StartFailed` case still picks between `Recording_UnknownError_Format(code)` and the bare `Recording_UnknownError`.

Better than I had it for a second reason: `ChatAudioControls` and the other recording controls can now get the same text from the status instead of repeating the mapping — which was the point of putting the status behind one compute method to begin with.

Re-verified all nine kinds through the new method against the running app; each still renders its own text and severity.

PR #4268, opened 2026-08-26
Video: iOS self-preview freeze — combined change set for review

alexyakunin:
**Review-only.** Every commit here is already on `dev`; this branch exists so the day's video work can be read as one diff. Every file it touches is byte-identical to `dev`, so merging it is a no-op.

Branched from `4dec4658ff` — the commit immediately before the first video change.

## What the day was

The iOS own-camera preview froze after a camera switch, two distinct ways: sometimes on a single frame, sometimes live-but-invisible. Capture, encoding and the outbound stream were healthy at 24fps throughout. Six hypotheses were tested and killed by measurement before instrumentation localised it; two of those became commits that are reverted here.

Both faults are [WebKit bug 230922](https://bugs.webkit.org/show_bug.cgi?id=230922) — open since iOS 15, still reproducing on 26, no code-level fix upstream.

## What actually ships

| | |
|---|---|
| `f03ee9c5e8` | Blazor's class-attribute diff was wiping JS-set classes. JS-owned render state moves to `data-*`. **A real, separately-verified bug.** |
| `fec8c4a267` | WebKit paints the self-preview on canvas instead of a `VideoTrackGenerator` track — avoids the bug and measured cheaper |
| `79cdb5c8a3` | Rotation for the join-modal preview, which the canvas painter needs and the generated track didn't |
| `f2a858fdb7` | Removes the stall detector and repair ladder |

Plus `forceRecomposite` (one `display` toggle on `loadeddata`) and preview instrumentation behind `isPreviewTraceEnabled()`, off by default.

## Dead ends, kept in history

`bc6eecf6d3` (writer-jam detector) and `e9bec8d1c6` (generator reuse) were both reverted — the first never fired, the second could not fire because `switchCamera` tears the worker down. They're here because the reverts are part of the story.

## CPU measurement

30s Activity Monitor captures on an iPhone 13 Pro, two runs per arm:

| process | MSTG | canvas |
|---|---|---|
| `WebKit.GPU` | 12.3% | 19.4% |
| `backboardd` | 22.0% | 12.8% |
| `WebKit.WebContent` | 31.0% | 28.9% |
| `ActualChat` | 8.6% | 6.6% |
| **subtotal** | **75.8%** | **69.3%** |

Painting moves into the GPU process, but the compositor stops handling a dedicated video layer and gives back more.

**Caveat:** the ladder was collapsed to 320×184 with no viewer subscribed, so canvas was measured at a quarter of the pixels a real call paints. Unverified at 640×360.

## Worth a look

- Whether canvas should stay the WebKit default given that caveat
- `forceRecomposite` — kept deliberately, though with WebKit on canvas it only matters under `video.debug.previewCanvas='0'`
- The instrumentation — dormant, but it's what made the diagnosis possible

https://claude.ai/code/session_01TJtnveuFUHm2pcpWChoJhn

### DIFF
--- a/src/dotnet/UI.Blazor.App/Components/JoinVideoCallModal/join-video-call-modal.css
+++ b/src/dotnet/UI.Blazor.App/Components/JoinVideoCallModal/join-video-call-modal.css
@@ -51,9 +51,9 @@
 /* Canvas-fallback backend (Edge/Firefox/iOS): RecorderPreviewView paints
    frames to the canvas, not the <video> (which has no track). Reveal the
    canvas as the primary surface — it sits after the <video> in the DOM, so it
-   covers the empty element. preview-backend-canvas is set by the view on
+   covers the empty element. data-preview-backend is set by the view on
    attach. */
-.join-video-call-modal .video-frame.preview-backend-canvas .camera-preview {
+.join-video-call-modal .video-frame[data-preview-backend="canvas"] .camera-preview {
     @apply block;
 }
 /* Mirror toggle — overrides the default scaleX(-1). Placed on .c-video (owned
--- a/src/dotnet/UI.Blazor.App/Components/JoinVideoCallModal/join-video-call-modal.css
+++ b/src/dotnet/UI.Blazor.App/Components/JoinVideoCallModal/join-video-call-modal.css
@@ -11,6 +11,13 @@
     @apply overflow-hidden;
     @apply bg-03;
 }
+/* Query container for the .rotated-video axis swap below, scoped to the canvas
+   painter so the generated-track path keeps its original layout untouched.
+   Safe to size-contain here: the surfaces inside are absolutely positioned, so
+   the frame's height comes from min-h-56 rather than from its content. */
+.join-video-call-modal .video-frame[data-preview-backend="canvas"] {
+    container-type: size;
+}
 .join-video-call-modal .video-frame .plug-text {
     @apply text-title-1 text-02;
 }
@@ -35,7 +42,8 @@
     @apply object-cover;
     @apply rounded-2xl;
     @apply hidden;
-    transform: scaleX(-1);
+    transform: scaleX(-1) rotate(var(--video-rotation, 0deg));
+    transform-origin: center center;
 }
 /* Native <video> visible when Join mode has a live track. Blur-on case swaps
    the <video>'s srcObject to the worker's MediaStreamTrackGenerator output,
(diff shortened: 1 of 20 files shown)


src/dotnet/UI.Blazor.App/Components/JoinVideoCallModal/join-video-call-modal.css
alexyakunin, line 45: It changes rotation on non-WebKit devices, right? Should it?

alexyakunin: For the context, video preview looked fine on Android, etc.

alexyakunin, line 72: Same here.

alexyakunin: You're right, and it's now fixed on `dev` in c800c8d650.

I'd reasoned it was a no-op — `rotate(var(--video-rotation, 0deg))` is the identity when nothing sets the variable, and on the generated-track path `preview-forwarder.ts` reports `displayRotation = 0` (Chromium stamps rotation as `VideoFrame` metadata, so the element rotates itself), which makes `applyRotationLayout` *remove* the variable.

But that only holds once frames are flowing. `applyRotationAndFit` falls back when there's no reported presentation yet:

```ts
const rotation = previewPresentation
    ? normalizeRotationQuarter(previewPresentation.rotation)
    : normalizeRotationQuarter(currentDelta - this.initialDeviceScreenDelta);
```

That delta is 0 at startup, but diverges when the screen orientation is locked and the device is physically rotated. In that window Android would have picked up a turn the modal never applied before — a real behaviour change, exactly as you suspected.

The rules now hang off `[data-preview-backend="canvas"]` rather than keying off the variable, so they can only match the canvas painter. Every pre-existing selector is untouched and the change is purely additive — the generated-track path is byte-identical to before on Android and every other non-WebKit engine.

alexyakunin: That context is what makes the fix the right one rather than a defensive tweak — Android looked fine because it *was* fine in the common case, and I shouldn't have put a rule in its path at all.

Worth spelling out why Android was unaffected in normal use: on Chromium `HAS_VF_ROTATION_INIT` is true, so the sender stamps display rotation as `VideoFrame` metadata and reports `displayRotation = 0` to the view — the `<video>` element does the turn itself and CSS is meant to stay out of it. The only gap was the pre-first-frame fallback to the device-vs-screen delta.

Now scoped to `[data-preview-backend="canvas"]`, so Android/Chromium can't match those rules at all.

alexyakunin: Same fix — this block is now `[data-preview-backend="canvas"] .camera-preview.rotated-video`, so it can only match the canvas painter.

Two extra notes on this one specifically:

- `.rotated-video` is only added for **odd** quarter-turns, so even before scoping it needed the fallback delta to land on 90°/270°. Narrower than the base rule, same root cause though.
- The `100cqw`/`100cqh` axis swap needs a container-query context, and `container-type: size` is already scoped to `[data-preview-backend="canvas"]` for the same reason — size containment is a real layout change and the generated-track path has no use for it.

PR #4288, opened 2026-08-29
perf(android): keep the main looper free during startup

alexyakunin:
## Why

Play Vitals flags the app as **"Bad behavior"** — user-perceived ANR rate **0.93%** against a 0.47% threshold, which suppresses discoverability. Crash rate (0.12%) is fine.

Over Aug 22–29 production had **zero crashes and 10 ANR clusters / 15 events**, in two families:

| # | Family | Stack | Events |
|---|---|---|---|
| 1 | Push-woken cold start | `MainApplication.onCreate` → `handleBindApplication`, RyuJIT frames on top (`fgFindBasicBlocks`, `allocateRegistersMinimal`, `isContained`, `emitOutputInstr`) | 13 |
| 2 | Launch-path UI thread | `BlazorAndroidWebView.<init>` → `WebViewChromium.init`, "Input dispatching timed out (No focused window)" | 2 |

An ANR is caused by the length of **one uninterrupted main-thread block**, not by total work — so this PR targets block length. It deliberately does not chase the ~20% generic-instantiation / cctor cost from the 2026-07 profile in `docs/android-specific.md`; that's a throughput problem, and that doc already records a failed attempt in that direction.

Two ideas that came up were already implemented and are left alone: the Blazor DI container is already lazy and built off the main thread, and `WarmUpWebView` already exists (and was deliberately kept out of `MainApplication.OnCreate`).

### Headless start skips UI-only init — family 1

`MauiStart` reads `RunningAppProcessInfo.Importance` through `AndroidUtils.GetProcessInfo()` (the accessor `GetProcessStartReason` already used) to tell a broadcast-started process from a user launch. `CreateMauiApp` then skips `WarmupStaticServices`, `SetupBlazorViewAppPostBuildRoutine`, `MarkAppBuilt`, `EnsureStarted` and the Chromium warm-up.

Safe because the container is **already built on demand by whoever needs it** — `PttWakeHandler.cs:54` calls `EnsureStarted()` itself and `PttSession` awaits `WhenAppReady`, while ordinary notification display touches Android APIs only. Detection is fail-safe: an unknown importance takes the full interactive path, so a broken check costs startup time, not correctness. An Activity reaching a headless process (a notification tap) calls `MauiProgram.PromoteToInteractive()`, which runs exactly what was skipped, guarded by an `Interlocked.Exchange` so a normal launch pays nothing.

### WebView attaches when ready instead of blocking — family 2

`CustomBlazorWebViewHandler.SetMauiContext` spin-waited on the main thread — `while (!WhenAppReady.IsCompleted) Thread.Sleep(5)` — for the entire container build, burning the core that finishes the very build it was waiting on, on top of Chromium's provider load.

`MainPage` now leaves `Content` null and attaches the WebView once both the container and the warm-up are ready, off the UI thread. Its `BackgroundColor` is already `SplashBackgroundColor`, so the native splash stands in and the window still draws and can take focus. `Content: null` was already an expected state (the foreground handler recreates from it), so that handler now also checks `IsWebViewAttachPending` to leave the initial attach alone. `MaxAttachDelay` caps the wait at 10s and falls back to today's behaviour rather than leaving a blank splash.

Both the container build and the warm-up now start at the end of `CreateMauiApp` on interactive starts, so they run alongside MAUI's own startup rather than beginning when the Activity already needs them.

### Diagnostics

`MauiSettings.Diagnostics.MainThreadBlockThreshold` now drives `AndroidMainThreadMonitor`. It stays **off by default** — any `Looper` printer feeds the GC pressure it is meant to observe — but the knob is discoverable for a measurement run.

## Diff noise

`MainActivity.cs` shows ~88 changed lines but the functional change is three added lines at the top of `OnCreate` plus the relocation of `WarmUpWebView` to `AndroidUtils`. The rest — moving `CallKeyguardDismissCallback` under `// Nested types`, sealing `SplashDelayer`, wrapping three long lines — is the style hook requiring pre-existing violations in touched files to be fixed. Same for the blank-line insertions in `MauiProgram.cs` and `AndroidUtils.cs`.

## Verification

Builds clean (`net11.0-android`, Debug): 0 errors; the one warning in a touched file (`BL0016`, `MauiProgram.cs:303`) is pre-existing.

**Not yet verified on device.** This touches notifications, which is where a "skip" breaks things silently, so before merge:

- [ ] Push to a headless process (`am kill`, never `am force-stop`) — notification still appears, for each `NotificationKind` branch: chat message, `IncomingCall`, `SpeechStarted` (PTT wake), `Attention`, and a dismissal (`DismissedTags`)
- [ ] Tap a notification on a headless process → promote produces a fully working app (sign-in intact, chat opens, audio works)
- [ ] Cold launcher tap, and warm start (home → relaunch, the `_isFirstTime` path)
- [ ] No WebView provider → still reaches `WebViewMissingActivity`
- [ ] `AndroidMainThreadMonitor` longest-single-block before/after, both start kinds
- [ ] Re-record `android-notif.mibc` — this changes which methods the push path runs, so the profile committed in `44967d77f5` goes stale on merge

Play Vitals ANR rate is a 28-day rolling metric, so the device-side block measurements are the acceptance criterion, not Vitals.

https://claude.ai/code/session_01Jc8rXaMZ9kAKNSVjHvFNZN

### DIFF
--- a/src/dotnet/App.Maui/Platforms/Android/AndroidUtils.cs
+++ b/src/dotnet/App.Maui/Platforms/Android/AndroidUtils.cs
@@ -17,6 +17,8 @@ public static class AndroidUtils
     private const int ImageDownloadTimeoutSeconds = 5;
     private const string Tag = Firebase.Messaging.Constants.Tag;
     private const string ThreadNetworkIo = "Firebase-Messaging-Network-Io";
+    private static readonly Lock WebViewWarmUpLock = new();
+    private static Task? _whenWebViewWarm;
 
     public static bool? IsAppForeground()
     {
@@ -62,6 +64,35 @@ public static class AndroidUtils
         return processInfo is null
             ? "Unknown"
             : $"{processInfo.Importance}/{processInfo.ImportanceReasonCode}";
+    }
+
+    public static Task WarmUpWebView()
+    {
+        // Takes Chromium's provider lock off the UI thread, which BlazorAndroidWebView's ctor
+        // would otherwise take on it. Never block the UI thread on the result: Chromium posts
+        // its native init back there, so waiting would deadlock.
+        // Volatile pair: the lock's release does not publish to a reader that skips the lock.
+        var whenWarm = Volatile.Read(ref _whenWebViewWarm);
+        if (whenWarm is not null)
+            return whenWarm;
+
+        lock (WebViewWarmUpLock) {
+            whenWarm = _whenWebViewWarm;
+            if (whenWarm is not null)
+                return whenWarm;
+
+            // Started inside the lock, so a lost race cannot leave a second provider load running.
+            whenWarm = Task.Run(() => {
+                try {
+                    _ = Android.Webkit.WebSettings.GetDefaultUserAgent(Application.Context);
+                }
+                catch (System.Exception e) {
+                    Android.Util.Log.Warn(MauiDiagnostics.LogTag, $"WebView warm-up failed: {e.Message}");
+                }
+            });
+            Volatile.Write(ref _whenWebViewWarm, whenWarm);
+            return whenWarm;
+        }
     }
 
     public static bool IsMainThread()
--- a/src/dotnet/App.Maui/Platforms/Android/AndroidUtils.cs
+++ b/src/dotnet/App.Maui/Platforms/Android/AndroidUtils.cs
@@ -32,7 +32,7 @@ public static class AndroidUtils
 
         // Screen is on and unlocked, now check if the process is in the foreground
 
-        int pid = Android.OS.Process.MyPid();
+        var pid = Android.OS.Process.MyPid();
         var appProcesses = activityManager.RunningAppProcesses;
         if (appProcesses != null) {
             foreach (var process in appProcesses) {
@@ -40,22 +40,28 @@ public static class AndroidUtils
                     return process.Importance == Importance.Foreground;
             }
         }
+
         return false;
+    }
+
+    public static ActivityManager.RunningAppProcessInfo? GetProcessInfo()
+    {
+        try {
+            var processInfo = new ActivityManager.RunningAppProcessInfo();
+            ActivityManager.GetMyMemoryState(processInfo);
+            return processInfo;
+        }
+        catch (System.Exception) {
+            return null;
+        }
     }
 
     public static string GetProcessStartReason()
     {
-        // Importance separates a broadcast-started process from a user launch: an FCM wake reads
-        // as a receiver/cached importance, a user launch as Foreground (AMS marks the process
-        // top-bound at bind time, before any activity exists).
-        try {
-            var processInfo = new ActivityManager.RunningAppProcessInfo();
-            ActivityManager.GetMyMemoryState(processInfo);
-            return $"{processInfo.Importance}/{processInfo.ImportanceReasonCode}";
-        }
-        catch (System.Exception e) {
-            return e.GetType().Name;
-        }
+        var processInfo = GetProcessInfo();
+        return processInfo is null
+            ? "Unknown"
+            : $"{processInfo.Importance}/{processInfo.ImportanceReasonCode}";
     }
 
     public static bool IsMainThread()
@@ -84,6 +90,7 @@ public static class AndroidUtils
             var networkIoExecutor = NewNetworkIoExecutor();
             imageDownload.Start(networkIoExecutor);
         }
+
         return imageDownload;
     }
 
@@ -125,6 +132,7 @@ public static class AndroidUtils
              * when it is updated.
              */
         }
+
         return null;
     }
 }
--- a/docs/android-specific.md
+++ b/docs/android-specific.md
@@ -13,6 +13,217 @@ that makes up the app packages.
 partial-R2R configuration and the measured numbers live. This page covers CPU profiling:
 where time goes at startup, and how to read the traces without being misled.
 
+## Startup: what runs where
+
+Startup is shaped by one constraint: an ANR is caused by the length of a **single
+uninterrupted main-thread block**, not by total work. So the goal is to keep the main
+looper free to dispatch, even when the same work still happens. Everything below follows
+from that.
+
+### Two kinds of start
+
+[`MauiStart`](https://github.com/Actual-Chat/actual-chat/blob/main/src/dotnet/App.Maui/MauiStartKind.cs)
+reads `RunningAppProcessInfo.Importance` through
+[`AndroidUtils.GetProcessInfo`](https://github.com/Actual-Chat/actual-chat/blob/main/src/dotnet/App.Maui/Platforms/Android/AndroidUtils.cs)
+and splits the process into two:
+
+| Kind | How it starts | Importance |
+|------|---------------|------------|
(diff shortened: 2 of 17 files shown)


alexyakunin:
## Device verification — Samsung SM-S948U1 (`m3q`), Android 16

Release build `2.18.41-alpha-g17f74ed6a2` (34734121), installed over the existing app (`-r`, no uninstall) so the sign-in was preserved.

### Passed

**1. Headless push to a killed process** (`am kill`, never `am force-stop`; chat message sent from the dev bot):

```
0.000s  -> CreateMauiApp
0.021s  -> MauiApp.CreateBuilder
0.033s  <- MauiApp.CreateBuilder — took 0.012s
        BlazorWebViewApp is initialized          ← factory stored, container NOT built
0.035s  <- CreateMauiApp — took 0.036s
        OnMessageReceived: kind 'Message'
        -> ShowChatMessageNotification
```

`CreateMauiApp` returns in **36 ms**, and no `Static services warmed up` / `MarkAppBuilt` / `start has been requested` appears anywhere before the tap — the headless skip does what it says. The notification was displayed (`notificationKey=0|chat.actual.dev.app|0|p-laP7rD-…`).

**2. Notification tap on that headless process** — `PromoteToInteractive` runs what was skipped, and the container builds off the main thread:

```
7.714s  MarkAppBuilt + "BlazorWebViewApp start has been requested"   ← promote
7.746s  base.OnCreate — took 0.030s        ← main thread free in 30 ms
7.803s  BlazorWebViewApp ready (bg thread 12830)
7.842s  MauiWebView Current = #1           ← WebView attached AFTER ready
8.508s  MarkFirstSplashRemoved
```

`MauiSession: Successfully read stored Session` — sign-in intact, chat opened from the notification link.

**3. Cold launcher tap** — the head start lands:

```
0.034s  MarkAppBuilt + "start has been requested"   ← now inside CreateMauiApp
0.035s  <- CreateMauiApp — took 0.035s
        BlazorWebViewApp ready (bg thread 14254)    ← ready before the Activity finishes
0.090s  base.OnCreate — took 0.031s
1.103s  MarkFirstSplashRemoved                      ← full UI
```

**4. Warm start** (home → relaunch): 27 ms, Activity reused, WebView untouched.

**5. `blocked the UI thread` warning fired 0 times** across every run — `SetMauiContext` always found the container already built, so the old `Thread.Sleep(5)` spin path was never entered.

**6. The `IsWebViewAttachPending` guard works.** On cold launch `OnBecameForeground` (OnStart) fires at ~0.09 s while `Content` is still null, but only one WebView is ever created (`Current = #1`). Without the guard OnStart would have raced a second one in.

**7. 0 FATAL / AndroidRuntime exceptions** in any run.

### Still not verified

- **Other `NotificationKind` branches** — `IncomingCall`, `SpeechStarted` (PTT wake), `Attention`, `DismissedTags`. The dev bot only produces `kind: 'Message'`, so these need a hand-crafted data message or a real call/PTT. `SpeechStarted` is the one worth care: it reaches `HeadlessBlazorScope.GetOrCreate()` → `BlazorWebViewApp.Current`, which is why `PttWakeHandler` calls `EnsureStarted()` itself — that call is untouched by this PR, but it is unexercised here.
- **No-WebView-provider path** (`WebViewMissingActivity`).
- **A before/after delta.** These are absolute numbers from the new build only; I did not measure the pre-change build on this device, so "0 UI-thread blocks" is meaningful on its own but is not a quoted improvement.

alexyakunin:
## PTT and dismissal verified

Confirmed on device by @alex-yakunin. Two more `NotificationKind` branches exercised on `2.18.41-alpha-g17f74ed6a2`:

**`SpeechStarted` (PTT wake)** — the branch that reaches `HeadlessBlazorScope.GetOrCreate()` → `BlazorWebViewApp.Current`, i.e. the one the headless skip could plausibly have broken:

```
OnMessageReceived: kind 'SpeechStarted', CollapseKey='speech-started-s-P7oRNDTeHM-VpxAh5qJ6t'
[HeadlessBlazorScope] Headless scope created
OnMessageReceived: kind 'Message' — "Voxt RoboKitty (Dev): Okay. Uh, wait, testing— Okay."
```

Full round trip: wake → headless scope → record → transcribe → post → notification back. No `Couldn't start scoped services`, no FATAL, no UI-thread block.

**`DismissedTags`** — dismissal push (`dismissedTags: 's-P7oRNDTeHM-VpxAh5qJ6t'`) cancelled the notification cleanly.

### Coverage now

| Branch | Status |
|---|---|
| `Message` | verified (headless cold start + warm) |
| `SpeechStarted` | verified (warm process) |
| `DismissedTags` | verified |
| `IncomingCall` | not tested |
| `Attention` | not tested |

### One caveat on the PTT test

It ran against a **warm** process (pid 14175, container already built), not a headless cold start. `am kill` could not kill it — with PTT armed, `AndroidActivitiesForegroundService` holds the process, and `am kill` only reaps background ones.

That is also why the cold-headless-PTT case is rare in production: an armed PTT process generally does not die. Reaching it needs a reboot. Risk assessed as low — `PttWakeHandler.cs:54` calls `BlazorWebViewApp.EnsureStarted()` before touching the scope, and this PR does not touch that line; it is precisely the on-demand build that makes the headless skip safe.

src/dotnet/App.Maui/Platforms/Android/AndroidUtils.cs
```diff
@@ -32,32 +34,55 @@ public static class AndroidUtils
...
+    }
+
+    public static Task WarmUpWebView()
+    {
+        // Without this the WebView provider is loaded on the UI thread inside performStart, right
+        // before BlazorAndroidWebView is constructed. GetDefaultUserAgent forces the same load and
+        // is safe off the UI thread. Nothing may block the UI thread on the returned task:
+        // Chromium posts its native init back to that thread, so waiting there would deadlock.
+        lock (WebViewWarmUpLock)
+            return _whenWebViewWarm ??= Task.Run(() => {
```
alexyakunin: Why won't you return _whenWebViewWarm pre-lock, if it's non-null?

alexyakunin: Done — added the pre-lock fast path.

It needs both barriers to be safe, though, because the fast path deliberately skips the lock and a lock’s release only orders things for a reader that *acquires* it:

- `Volatile.Read` (acquire) — stops later loads floating above the reference load.
- `Volatile.Write` (release) — guarantees the Task’s construction is visible before the reference goes non-null. Without it an ARM64 reader can see non-null with stale internals.

On x64 both are a plain `mov`; on ARM64 it is `stlr`/`ldar` instead of `str`/`ldr` — one instruction, no fence. This is also the rule in `docs/CODING_STYLE.md`: "A `lock` around the write is not a substitute for the release. Publishing a newly built object under a lock still needs `Volatile.Write`."

In fairness: `Task.Run` does interlocked work to queue the item, so the store-store ordering is probably already there in practice — but that is an implementation detail of `Task.Run`, not a guarantee.

The `Task.Run` still starts *inside* the lock, so a lost race cannot leave a second Chromium provider load running — the bug in the original CAS version you caught.

PR #4290, opened 2026-08-30
fix(audio): iOS loudspeaker routing, and the silent-stall family behind it

alexyakunin:
Fixes the iOS earpiece-routing bug, then everything a five-agent review of the audio pipeline turned up behind it. Each fix is its own commit.

## The bug that started this

Incoming voice played out of the **ear speaker** whenever you were the one who started recording. Joining a chat where someone was already talking was fine — which pointed at ordering, not configuration, since both cases end with an identical session.

It's a known iOS defect ([Apple forum 721535](https://developer.apple.com/forums/thread/721535)): once an `AVAudioEngine` with VoiceProcessingIO is running, **any source started afterwards plays on the receiver while the session keeps reporting the speaker**. Measured on an iPhone 13 Pro: 33 consecutive `ApplyOutputRoute` calls logged `Speaker -> Speaker -> Speaker` while the audio was audibly on the earpiece. Restating an override the session already holds is a no-op, so the fix has to make it a *transition* — clear to `.none`, then set `.speaker` — re-applied per source, not once per recording.

Two things found on the way: the session mode was `Default`, which `SetVoiceProcessingEnabled` silently replaces with `VoiceChat` (taking `DefaultToSpeaker` with it) and which also **disables VPIO's echo cancellation outright** — so the app was running without AEC. And the recording engine had no output bus, so its downlink DSP errored on every render: 11,004 `render err: -1` lines in one 35k-line capture. Both fixed.

Verified on device.

## Probable cause of "streams stop until the sender restarts"

Independently confirmed in both senders, .NET and TypeScript, with the same incorrect comment in each:

The push retry loop resets `frameIndex` but **reuses the original stream start**. A retry within `MaxBeginsAtDrift` (5 s) therefore registers a new streamId whose `BeginsAt` *equals* the dead one's — and `ListeningStreamMuxer` replaced a same-author stream only on a **strictly** greater `BeginsAt`. So it cancelled the new live stream, kept the dead one, and nothing re-scans afterwards. Listeners go silent for the rest of the utterance while the server keeps ingesting and transcribing — which is why the server looks healthy and live captions keep scrolling over dead audio.

Fixed on all three sides: both senders advance the claimed start, and the muxer gives ties to the newcomer.

## Everything else

**Cross-platform** — a full send channel dropped and leaked the packet instead of retrying it (found independently by two reviewers). The Opus codec caught only cancellation and then completed its channel *without* the exception, so one bad packet reported a clean end mid-utterance with no log line anywhere. `TrackPlayer` kept feeding an engine that had already reported its end, and `AudioTrackPlayer` swallowed every buffer-low timeout — degrading to one frame per 10 s forever rather than failing.

**iOS** — tunes and `Resume()` never restated the route (the recording-confirm tune hit this on *every* recording). A disposed scope could be handed back to the next consumer, tearing down the recording engine under a live capture. `_isSuspended` could stick and disable configuration-change handling for good.

**Android** — negative `AudioTrack.Write` results were ignored entirely, so a dead track silently discarded the rest of the stream. A dead capture producer left the consumer parked forever, behind a comment that already described the wedge.

**Windows** — the playback graph subscribed to no error or device events. Tracks under 100 ms never started and hot-spun a core. A throw in the capture loop left the mic hot with nothing recorded. `RecordingStopped` was unsubscribed, which is the *only* channel NAudio has for a capture-thread failure, so losing the render endpoint silently ended echo cancellation. A loopback that couldn't start killed the whole recording. And AGC was driving a mic that wasn't the one being captured.

**Web** — resampling threw `RangeError` before it ever ran on any non-48 kHz device (silent sender on Firefox), and the VAD pushed each window twice. A nulled audio stream could never be recovered. A failed decoder flush suppressed the track end, hanging the track forever. One rejection poisoned a worklet's promise chain permanently.

## The last round

Everything remaining from the reviews, after the items above:

**Cross-platform** — `_encodingFrameSourceCapturedAts` was a plain `Queue<Moment>` written from the capture task and read from the encode worker; it's concurrent now. `TrackPlayer`'s `EndCommand` and the engine's `DisposeAsync` were the only unbounded waits on the normal teardown path, and `Playback.OnAbortCommand` awaits them inline in its command loop — so one hung engine call stopped listening being started or stopped anywhere in the app. Both bounded. `BlockRingBuffer.Dispose` returned its pooled array outside the lock while `TryWrite`/`TryRead` copy under it — a producer mid-write kept writing into an array the pool had already handed to someone else, and that pool backs every audio buffer in the app.

**Server** — a transient mid-relay failure put the streamId in `_excludedStreamIds` for the life of the subscription, silencing that speaker until they started a new stream. It resumes now, bounded, and is marked preexisting so `GetSkipTo` serves it at the live edge rather than replaying the utterance from zero.

**Android** — focus-restore handlers were only ever appended, so the Nth recover replayed all N cycles' restores. `Mode.InCommunication` is raised before the focus request and was left stranded on denial — the normal answer during a phone call, which is exactly when a PTT wake arrives. Plus an `ObjectDisposedException` race in `WhenUnpaused` that killed live tracks on every focus churn, and a tune's focus scope leaking when playback threw.

**Windows** — graph-creation failures now log (every layer above swallows them, so a process-wide `CreateAsync` failure showed up as every track ending normally with no trace); the playback quantum was computed from the recording rate; `endpointVolume` was never disposed.

**Web** — `streamErrorDelayMs` exists for the push retry loop and was referenced nowhere, so a persistent rejection spun one call per round trip; `AsyncProcessor` left `_isRunning` true after its loop died, so `enqueue` kept accepting items nothing would process.

## Not fixed — deliberate

- **Focus loss ends live listening and doesn't resume it.** Confirmed intentional; `docs/live-audio/07-receiver.md` promised the opposite and was corrected instead.
- **`WhenPlaybackDrained` doesn't complete for a paused track.** Correct as written — a paused track shouldn't end — and the abort path plus the newly bounded `End` cap the exposure.
- **No render-default watcher on Windows (M5).** The serious case, the endpoint disappearing, is now covered by `UnrecoverableErrorOccurred`; the remaining harm is split output for at most one utterance.

One reviewer finding was **rejected as a false positive**: iOS `GetDiagnostics` reading `_activeScopes` without the lock. Both accessors only enumerate `_byMode`, which is fixed at construction and never mutated — they read an `int` Count, so no throw is possible.

## Verification

Three adversarial review rounds over the branch's **end state**, not per commit. They found **15 defects in my own fixes** — 6 high-severity in round 1, 2 in round 2, 1 in round 3 — all now fixed and re-verified. Worth knowing what they were, because the pattern is instructive:

- A `Cancel()` on a CTS the other side disposes — throws, and skipped the rest of a `finally` on the **normal** stop path.
- A `TryRemove` that removes before comparing identity, so my "drop the stale scope" fix evicted the *live* one every time it fired.
- A buffer-low escalation that couldn't tell a stopped engine from a **paused** one, killing any track paused over 30 s.
- A retry backoff that ran on the **success** path, delaying every following utterance and risking silent drops at `MaxStreams`.
- A "bounded" retry whose counter was wiped by a trailing `else` it fell through to.
- A disposed-buffer read returning a **completed** task, spinning a core.
- An interruption guard that read a raw latch with no expiry — trading a 10 s glitch for a capture wedged until manually stopped, because the timeout it suppressed *was* the self-heal.
- An `AudioContextAction` retry that re-ran against the same dead context, where the recorder's early-out matched dead-against-dead and exited the loop permanently — reproducing the exact bug the loop was added to fix.

Two findings were **rejected**: an iOS `GetDiagnostics` lock (the accessors only enumerate a dictionary fixed at construction — no throw possible), and applying the 10 s interruption expiry inside the capture guard (a real phone call holds that latch for minutes; a 60 s deferral cap is the right trade, and the reviewer agreed).

Knowingly accepted, not overlooked: a stale `State.IsPaused` edge; supersede-exclusion silencing a genuinely-live superseded stream (the pre-fix comeback replayed from position 0, so this is the better of two imperfect behaviours); Android restoring `Mode.Normal` rather than the pre-raise mode; and the web retry backoff being 1 s against MAUI's 100 ms — a one-constant change if you'd rather have parity.

## Testing

All four .NET targets build (CI filter, android, windows, ios) and `npm run build:Verify` passes.

Full branch on device (iPhone 13 Pro): recording ran to the app's own 30 s silence auto-stop (`RecordChat: idle threshold reached`) with 0 ObjC exceptions, 0 recording failures, 0 `render err: -1`, and 0 stall-timeout firings — that last one matters, since the capture timeout had to *not* fire during normal recording or during the idle stop.

The iOS routing fix is confirmed against a real reproduction. Everything else is verified by build, by device soak, and by reading — not by reproducing each failure.

**The riskiest change is `AudioContextAction` re-running after a context failure.** It is shared by every web audio consumer and I could not reproduce a context failure to test it. Worth the closest look in review.

https://claude.ai/code/session_01G86jKcZ7xu1NL4y1fXAAQo

### DIFF
--- a/src/dotnet/App.Maui/MaciOS/Audio/AppleAudioCapture.cs
+++ b/src/dotnet/App.Maui/MaciOS/Audio/AppleAudioCapture.cs
@@ -10,6 +10,10 @@ public class AppleAudioCapture(AppUIHub hub) : IAudioCapture
 {
     private static readonly TimeSpan InputNodeHoldTimeout = TimeSpan.FromMinutes(30);
     private static readonly TimeSpan InputNodeHeartbeatTimeout = TimeSpan.FromSeconds(30);
+    // A tap delivers continuously regardless of speech, so a gap this long isn't silence.
+    private static readonly TimeSpan CaptureStallTimeout = TimeSpan.FromSeconds(5);
+    // Past InterruptionEndTimeout, so a Began that never gets its End can't hold this open.
+    private static readonly TimeSpan MaxInterruptionDeferral = TimeSpan.FromSeconds(60);
 
     private static long _inputNodeHeldAt;
     private static long _inputNodeHeartbeatAt;
@@ -47,7 +51,8 @@ public class AppleAudioCapture(AppUIHub hub) : IAudioCapture
         // their iterator: here the whole AVAudioEngine setup is inside CaptureInternal.
         => Task.FromResult(AudioCaptureResult.Ok(CaptureInternal(cancellationToken)));
 
-    private async IAsyncEnumerable<IMemoryOwner<float>> CaptureInternal([EnumeratorCancellation] CancellationToken cancellationToken)
+    private async IAsyncEnumerable<IMemoryOwner<float>> CaptureInternal(
+        [EnumeratorCancellation] CancellationToken cancellationToken)
     {
         // First statement, before TryTake() stops the pre-roll engine: everything below is
         // blocking native work, and a press landing in that window would see a free input node
@@ -119,13 +124,38 @@ public class AppleAudioCapture(AppUIHub hub) : IAudioCapture
             await AudioFocusUI.EnsureOutputRoute(cancellationToken).ConfigureAwait(false);
 
             var frameLen = Constants.Audio.OpusFrameLength;
+            var deferringSince = default(CpuTimestamp?);
             while (!cancellationToken.IsCancellationRequested) {
                 var owner = ArrayPools.SharedFloatPool.LeaseArrayOwner(frameLen, true);
                 if (!outBuffer.TryRead(owner.Span, out var whenReady)) {
                     owner.Dispose();
-                    await whenReady.WaitAsync(cancellationToken).ConfigureAwait(false);
-                    continue;
+                    if (await whenReady.TryWaitAsync(CaptureStallTimeout, cancellationToken).ConfigureAwait(false))
+                        continue;
+
+                    // hwFormat, the resampler and the tap are pinned at start, and Reconnect()
+                    // rebuilds none of them - so a route change that moves the mic (AirPods swap
+                    // 48kHz built-in for 8/16kHz HFP) leaves them built for a format that never
+                    // arrives, silently. Ending lets PushRecordingState rebuild the lot.
+                    cancellationToken.ThrowIfCancellationRequested();
+                    // An interruption stops the tap legitimately and RecoverInternal resumes the
+                    // engine after it, so ending here would split the recording. Bounded, though:
+                    // the flag is a raw latch with no expiry outside SetModeUnsafe, and iOS
+                    // doesn't guarantee an End.
+                    var focus = AudioFocusUI.GetDiagnostics();
+                    var isDeferred = focus is { IsInterrupted: true } or { IsSuspended: true };
+                    if (isDeferred) {
+                        deferringSince ??= CpuTimestamp.Now;
+                        if (deferringSince.GetValueOrDefault().Elapsed < MaxInterruptionDeferral)
+                            continue;
+                    }
+
+                    Log.LogWarning("CaptureInternal: no samples for {Elapsed} ({Reason}) - ending the capture",
+                        (deferringSince?.Elapsed ?? CaptureStallTimeout).ToShortString(),
+                        isDeferred ? "still interrupted" : "input stopped");
+                    yield break;
                 }
+
+                deferringSince = null;
                 yield return owner;
             }
             yield break;
@@ -136,7 +166,8 @@ public class AppleAudioCapture(AppUIHub hub) : IAudioCapture
                 // The engine is alive and still owns the input node - see IsInputNodeHeld.
                 Volatile.Write(ref _inputNodeHeartbeatAt, CpuTimestamp.Now.Value);
                 try {
-                    var estimatedResampledLength = pcmBuffer.FrameLength / hwFormat.SampleRate * AudioEngine.VoiceRecordingFormat.SampleRate;
+                    var estimatedResampledLength = pcmBuffer.FrameLength / hwFormat.SampleRate
+                        * AudioEngine.VoiceRecordingFormat.SampleRate;
                     if (outBuffer.RemainingCapacity < estimatedResampledLength) {
                         Log.LogWarning("Buffer full, dropping samples");
                         return;
--- a/src/dotnet/App.Maui/MaciOS/Audio/AppleAudioPlaybackEngine.cs
+++ b/src/dotnet/App.Maui/MaciOS/Audio/AppleAudioPlaybackEngine.cs
@@ -7,7 +7,7 @@ using ActualLab.Opus.MaciOS;
 
 namespace ActualChat.App.Maui.Audio;
 
-public class AppleAudioPlaybackEngine(
+public sealed class AppleAudioPlaybackEngine(
     string playerId,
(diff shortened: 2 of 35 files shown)


src/dotnet/App.Maui/MaciOS/Audio/AppleAudioCapture.cs
```diff
@@ -119,24 +126,66 @@ private async IAsyncEnumerable<IMemoryOwner<float>> CaptureInternal([EnumeratorC
...
+                        isDeferred ? "still interrupted" : "input stopped");
+                    yield break;
                 }
+
+                deferringSince = null;
                 yield return owner;
             }
             yield break;
 
+            static async Task<bool> TryWait(Task whenReady, CancellationToken cancellationToken)
```
alexyakunin: Convert this to TryWaitAsync extension method

src/dotnet/App.Maui/MaciOS/Audio/AppleAudioPlaybackEngine.cs
```diff
@@ -67,11 +66,13 @@ public Task Pause(CancellationToken cancellationToken)
         return Task.CompletedTask;
     }
 
-    public Task Resume(CancellationToken cancellationToken)
+    public async Task Resume(CancellationToken cancellationToken)
```
alexyakunin: Probably doesn't need to be async?

src/dotnet/App.Maui/Platforms/Android/Audio/AndroidAudioCapture.cs
```diff
@@ -68,6 +68,9 @@ public Task<AudioCaptureResult> Capture(CancellationToken cancellationToken)
         }
 
         var buffer = new BlockRingBuffer<float>(Constants.Audio.RecordingSampleRate * 10);
+        // The consumer parks on the ring buffer with no timeout, so a producer that dies has to
+        // say so - otherwise the whole recording wedges behind a wait nothing will ever satisfy.
+        var whenProducerEndedSource = new CancellationTokenSource();
```
alexyakunin: Rename to whenProducerEndedCts

src/dotnet/App.Maui/Platforms/Android/Audio/AndroidAudioCapture.cs
```diff
@@ -222,21 +228,40 @@ void Producer()
 
         async IAsyncEnumerable<IMemoryOwner<float>> Enumerate([EnumeratorCancellation] CancellationToken ct)
         {
+            using var endCts = CancellationTokenSource.CreateLinkedTokenSource(
```
alexyakunin: rename to cts or commonCts.

src/dotnet/App.Maui/Platforms/Android/Audio/AndroidAudioCapture.cs
```diff
@@ -222,21 +228,40 @@ void Producer()
...
                     yield return owner;
                 }
             }
             finally {
+                whenProducerEndedSource.DisposeSilently();
                 buffer.DisposeSilently();
             }
         }
+
+        static async Task<bool> TryWait(Task whenReady, CancellationToken ct)
```
alexyakunin: This should go away with TryWaitAsync extension

alexyakunin: (and this means the extension needs a couple of overloads)

PR #4299, opened 2026-08-30
Client-side partitioned command queue with optimistic reactions

crui3er:
A client command that can't reach the server right now — a reaction tapped on a
flaky connection, a read position written while the socket reconnects — used to
either fail in place or race the commands next to it. This branch gives such
commands a lane: they queue per partition key, retry until they land, and stay
visible while they wait.

## What's here

**`PartitionedCommandQueue`** (Core) is the ordering rule and nothing else: one
item per partition key runs at a time, different partitions run in parallel, and
it hands the caller what to run rather than running it. Keeping it a coordinator
rather than a worker is what lets the dispatching call stay the one that executes
the command, so context, cancellation and result all survive.

**`ClientCommandHandler`** is the command filter that puts this to work on clients.
Every `IQueuedCommand` goes into its lane and is re-dispatched when its turn
comes; transient failures are retried; every command's stage
(`Waiting/Running/Retrying/Completed/Failed`) is observable.

**Optimistic reactions** are a fold, not a hand-written patch: the badge shows the
server's summaries plus the reaction commands still in the queue, mirroring what
`ReactionsBackend.OnReact` would do with them. It converges by construction when
the queue drains, and clears by reconciliation rather than on a timer.

**Diagnostics**: `/test/command-queue` lists every tracked command live, and
`debugUI.suspendCommandQueue(suspend)` pauses the queue so commands pile up as
they would with no connection.

Two commands opt in so far: `Reactions_React` (no coalescing — a toggle must never
be collapsed) and `ChatPositions_Set` (`ReplaceWaiting` — only the newest read
position matters). `Chats_UpsertEntry` deliberately does **not**: message sending
already has its own per-chat lanes plus a persistent request repo that survives an
app restart, which is stronger than this queue offers. A command that declares no
partition key at all gets the retries and the registry but no lane, so it is never
serialized behind unrelated work.

## Three defects worth knowing about

Each was invisible in review and only surfaced under a real dispatch. All three
are fixed here and pinned by tests.

1. **The re-dispatch marker can't be an `AsyncLocal`.** `Commander.Run` wraps an
   outermost command in `TrySuppressFlow()` + `Task.Run` precisely so AsyncLocals
   don't flow into it, so the flag always read false on re-entry and the filter
   re-queued the command forever.
2. **The filter must outrank `ApiCommandDeduplicator`.** Below it, the outer
   dispatch claims the `Uuid` and then waits for a re-dispatch the deduplicator
   holds on that very claim — a deadlock that cleared only after the 5-minute TTL.
3. **The handler must be a singleton.** `CommandContext` creates a fresh DI scope per
   outermost command and resolves handlers from it, so a scoped handler handed every
   re-dispatch an empty instance, which didn't recognize the command as its own
   and queued it again — an unbounded loop that saturated the single WASM thread.
   The visible symptom was unrelated-looking: after one message the send button
   stopped working, in every chat except `Notes` (single-author chats don't write
   read positions, so they produce no queued command at all).

## Testing

- All unit tests pass after a rebase onto `dev` (751 in Chat.UI.Blazor.UnitTests,
  9 for the queue primitive), including 13 for the handler, 7 for the reaction
  fold, and `DiRegisteredQueueMustSurviveThePerCommandScope`, which fails on the
  pre-fix registration.
- Verified in the browser on a WASM client: two messages in a row send, the queue
  holds `ChatPositions_Set` in `Waiting` while paused, and Resume drains it to
  `Completed`.
- Found and fixed in the browser afterwards: taking away the only reaction left
  an empty row holding just its add button, because "is there anything to show"
  was answered from the queue rather than from the folded model.

## Before merging

One commit carries the superpowers spec/plan — `/prepare-merge` drops those.

https://claude.ai/code/session_01FWbM4RPMdDMg1iCobTR36s


alexyakunin:
Yes, yes!!!

PR #4302, opened 2026-08-30
fix(audio): stop the recording toggle turning itself off

alexyakunin:
## The bug

On the **Voxt Windows app**, tapping the mic button in the message editor while
**someone else is already talking** turns recording on and then instantly off
again — silently, no toast, no modal, no error. It works when nobody is talking,
it works if you start recording *first* and someone talks after, and it works in
a browser on the same machine.

The toggle's state is `ActiveChats[…].IsRecording` — pure intent — so an engine
failure can't turn it off (it leaves the button on and `PushRecordingState`
retries). Only `SetRecordingChatId(null)` can, and exactly one of its call sites
is silent, instant and uncorrelated with a tap.

## Root cause

`ChatAudioUI.RecordChat` picks its stop reason with `Task.WhenAny`, which returns
the first task to complete **in any state**, and builds `whenIdle` with the
2-arg `ForegroundTask.Run` overload that turns a throw into a *faulted task
without logging*. The winner was then deliberately not awaited:

```csharp
whenWinner = await Task.WhenAny(whenStopped, whenIdle, whenRecorderStopped);
// No need to await for the result of WhenAny: we're stopping anyway
…
if (ReferenceEquals(whenWinner, whenIdle)) { Log…("idle threshold reached"); await SetRecordingChatId(null); }
```

So **any** exception inside the idle observer was relabelled as an idle timeout,
cleared the intent, and left nothing behind but a misleading info-level log line.
A genuine idle stop takes 30s and shows a countdown, so an *instant* one was
always a fault.

The playback-conditionality comes from `ObserveStreamingIdleBoundaries`: its
`isHeld` branch is taken only while someone is streaming, and that branch loops
on `WhenInvalidated` → `Update()` → the `LiveSessions.HasActivity` RPC,
repeatedly, from the moment the mic opens. The idle branch instead parks for
~20s and touches nothing.

The same file already knew better — `StopListeningWhenIdle` carries the comment
*"Task.WhenAny itself never throws, so the winner must be awaited"*.

## Everything fixed

Verified by an independent audit pass before any code changed.

| | Fix |
|---|---|
| **A1** | `RecordChat` observes the winner and logs a faulted watcher; only a *successful* `whenIdle` clears the intent |
| **A2** | The idle observer reads `HasError` before `.Value`; an unreadable hold counts as held |
| **A3** | The Windows default-capture-device watcher compares against the endpoint the input node opened, so a no-op re-pick no longer ends the capture |
| **N4** | `WindowsAudioPlaybackEngine` stops its `AudioGraph` before disposing it — a graph left running wedges WinRT audio process-wide and kills every later `CreateAsync`, capture included |
| **B1** | `recorder-toggle.ts` / `playback-toggle.ts` used a static one-shot that bound only the buttons present at first render; both now share one delegated capturing listener |
| **B2** | `ActiveChatsUI.FixActiveChats` NRE'd on a recording chat without write access, aborting the whole update |
| **B3** | `startMicrophoneStream` rethrows instead of swallowing, so `state` and the heartbeat don't stay live for a pipeline with no mic |
| **B4** | The web heartbeat skips while the recorder is `'starting'` — it used to stop the start it was racing |
| **B5** | `MarkStarting` no longer carries the previous session's `IsRecording` into the new state |
| **B6** | `whenStopped` / `whenRecorderStopped` faults are logged rather than silently costing a restart cycle |
| **N1** | `SetRecordingChatId` restored keep-listening chats by mutating an array the updater had already returned — the write was lost. Now runs after the update, via the helper `InitializeListening` shares |
| **N2** | `LiveSessionUI`'s loops read `.Value` off possibly-errored computeds under a bare `Task.WhenAll`, so one error ended participation reporting *and* mute enforcement permanently. Now `AsyncChain` + `RetryForever`, and mute enforcement re-confirms after a settle delay so a peer's own mute lift doesn't race it |
| **N3** | `PushRecordingState` no longer replays the begin-recording tune on each restart |

Not changed: **A5** (the 500ms tap guard) and **A6** (the pre-`9784b72705`
loopback→`NoPermission` chain) — audited and found not to be defects here.

## Verified

- `npm run build:Debug` (`tsc --noEmit` + bundle) and the full server `dotnet build`: clean.
- `dotnet build App.Maui -f net11.0-windows10.0.22621.0`: 0 errors.
- In live Chrome against the dev server: a `div.recorder-wrapper > button` created
  *after* `init()` ran fires both `initContextInteractively` calls, and a
  non-matching button fires neither — the B1 regression, directly exercised.

https://claude.ai/code/session_01Rg8JuS6YcbvqAgYkDj8PRA

### DIFF
--- a/src/dotnet/UI.Blazor.App/Services/LiveSessionUI.cs
+++ b/src/dotnet/UI.Blazor.App/Services/LiveSessionUI.cs
@@ -18,6 +18,9 @@ public class LiveSessionUI(AppUIHub hub) : UIWorkerBase<AppUIHub>(hub), ICompute
     private static readonly TimeSpan HeartbeatInterval = TimeSpan.FromSeconds(45);
     // Give up watching if the call we just started never shows up as dialing.
     private static readonly TimeSpan DialingWaitTimeout = TimeSpan.FromSeconds(15);
+    // How long a mute verdict must survive before it stops a recording - long enough for a
+    // peer's own mute lift to come back from the server, short enough to feel immediate.
+    private static readonly TimeSpan MuteEnforcementDelay = TimeSpan.FromSeconds(1);
 
     private static readonly string JSStartRingback = $"{BlazorUIAppModule.ImportName}.OutgoingCallRingback.start";
     private static readonly string JSStopRingback = $"{BlazorUIAppModule.ImportName}.OutgoingCallRingback.stop";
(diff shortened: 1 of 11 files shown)


src/dotnet/UI.Blazor.App/Services/LiveSessionUI.cs
```diff
@@ -196,23 +219,37 @@ private async Task RunParticipationSync(CancellationToken cancellationToken)
...
             }
 
             await cMuted.WhenInvalidated(cancellationToken).ConfigureAwait(false);
             cMuted = await cMuted.Update(cancellationToken).ConfigureAwait(false);
         }
+        return;
+
+        // An errored computed must not reach .Value: the throw would take down the whole worker.
+        static bool IsMuted(Computed<ChatId?> computed)
+            => !computed.HasError
```
alexyakunin: Use .ValueOrDefault or destructuring.

PR #4303, opened 2026-08-30
fix(video): new codecs, new profile selection, fixes

alexyakunin:
## Why

A user on Firefox/Linux (CachyOS, Firefox 154.0.1) reported frozen video with unusually good diagnostics. He ruled out the obvious causes himself — FFmpeg 9.0.1, x264 and OpenH264 2.6.0 installed, and `ffmpeg -c:v libx264 -profile:v main -level:v 4.1` encoding 1280×720@30 faster than real time. **The OS had a working H.264 encoder throughout.**

```text
VideoEncoder.isConfigSupported({codec: "avc1.4D4029", ...}) -> supported: true
[VideoPipeline] encoder error (layer=0, codec=avc1.4D4029, 320x184): Operation is not supported
[VideoPipeline] Recorder pipeline failed: [ENCODER_INIT_FAILED] codec=avc1.4D4029
RecorderWorker: encoder init failure for codec=avc1.4D4029 (category=h264) - excluding and re-picking
repickCodecAndRestart: re-pick returned same codec avc1.4D4029
```

The first half is [Mozilla Bug 1918769](https://bugzilla.mozilla.org/show_bug.cgi?id=1918769) (UNCONFIRMED, unassigned, P3), reported on **Windows and macOS as well as Linux**. Firefox's OpenH264 GMP plugin is WebRTC-only and isn't wired to WebCodecs, so no codec install fixes it.

The second half is ours, and it turned out to be one of six defects. Two of them produce that same frozen tile on a browser that behaves correctly — so this is not a Firefox workaround.

**The through-line: we advertise a profile we never probed, then give both sides one chance to accept it.**

## What changed

One commit per fix. Plan: `docs/plans/h264-codec-selection-fixes.md`.

| | Fix |
|---|---|
| **1** | `excludeEncoderCodec('h264')` returns early — H.264 is the universal fallback, so dropping it wholesale leaves nothing. That made `ENCODER_INIT_FAILED` a no-op for H.264 and looped forever. Now the exact **profile** is excluded as well as the category, so a failed `avc1.640028` still leaves Main and Constrained Baseline. |
| **2** | Detection probed `avc1.4D401F` (Main 3.1) and reported `avc1.640028` (High 4.0), on the theory that "a working low-level profile implies higher ones". Main → High is a profile change, not a level step, and 3.1 → 4.0 raises max MB rate and DPB size. Now probes a best-first ladder and reports the first that passes. |
| **3** | `getDefaultCodec` short-circuited Firefox to Main 3.1 while `getCodecForCategory` gave it Main 4.1. Deleted (redundant). Its second-order effect — `avc1.4D401F` is never in `supportedCodecs`, so the exact-string lookup always missed and reported Firefox as software whatever the truth — is fixed by a category fallback in `findCodecInfo`. |
| **4** | The receiver built **one** H.264 candidate (HEVC gets eight) and only ever probed `prefer-hardware`. Now retries every candidate with `no-preference`, and threads the chosen acceleration into `initialDecoderConfig` (which previously hard-coded `prefer-hardware` and would have re-failed the configure). H.264 candidates widen only — see below. |
| **5** | The stream declares the top tier's codec for a three-encoder ladder. Safe today (all layers request the same string; divergence can only be a browser downgrade, and a decoder configured above the bitstream decodes it), so this is log-only. |
| **6** | Bare `h264`/`avc1` and the unknown fallback mapped to High 4.0 — the least portable of the three profiles. Now Constrained Baseline. |

### The one direction that matters

H.264 candidates **only ever widen**. Our own HEVC code documents why: Chrome's `isConfigSupported` doesn't cross-check the codec-string level against the description, so declaring *less* than the bitstream carries makes `configure()` succeed and `decode()` drop chunks **silently** — a frozen tile with no error. Over-declaring is safe. There is a property test over all three profiles we emit asserting no candidate is ever lower.

## Also in here

- **Deleted `webcodecs-encoder.ts` + `webcodecs-decoder.ts`** (~1000 lines). Orphaned — they reference only each other; the live path is `adapters.ts` + `video-decoder-bridge.ts`.
- **`style(video)` commit** normalizing `hevc-codec-selection.ts` to the guide (pre-existing violations the hook surfaced). Two long comment blocks kept and recorded in `.claude/style-bypasses.md` — both document silent upstream browser bugs, so trimming them is how the ladder gets shortened and the bugs come back. Approved by Alex.
- **Docs:** `03-codecs-and-layers.md` described scalability-mode probing and a `scalabilityMode` field on `LayerConfig`. Neither exists — temporal SVC was *removed* in `3ae12d7f8`, not disabled. Corrected, along with `07-receiver.md`.

## Verification

`tsc --noEmit`, `lint:all`, and 558 unit tests pass (18 new). Field verification on Firefox/Linux is still outstanding.

## Not in scope

The plan's "Wider codec strategy" section records two open questions this branch does **not** answer: whether WebKit supports VP8/VP9 in WebCodecs (decisive for whether VP9 can be a second floor — Apple ships no VP9 hardware, and sources contradict each other), and whether the pipeline runs on WebKitGTK at all, which the planned MAUI Linux backend needs. It also records why the libav.js wasm-fallback plan was dropped: the premise was wrong, and shipping a wasm codec would reclassify Voxt as an AVC codec-unit distributor.

Deferred: the Firefox `MediaStreamTrackProcessor` transfer failure (silently downgrades capture to the rVFC pump on every Firefox session) and the 30–40 s Chromium/Linux startup delay.

https://claude.ai/code/session_01P9bqJidRkGDe6YrHVertp5


src/dotnet/UI.Blazor.App/Components/VideoPanel/video-recorder.ts
```diff
@@ -2085,56 +2103,98 @@ export class VideoRecorder {
...
+
+            // rVFC alone does not carry this element on Firefox: it fires a
+            // handful of times over a minute instead of once per source frame,
+            // so capture starves and the stall surfaces as an encoder fault.
+            // A silence threshold can't catch that — the gaps are irregular
+            // rather than absent — so the timer runs unconditionally next to
+            // rVFC and `lastPushedMediaTime` keeps whichever fires second from
+            // resubmitting a frame. Chromium and WebKit take the MSTP path and
+            // never reach this branch.
+            const timerPumpPeriodMs = Math.max(1, Math.round(
```
alexyakunin, line 2203: Do we need that, or it was an artifact of very slow encoder (~20 frames queue, etc.)

src/dotnet/UI.Blazor.App/Services/Video/codec-support.ts
```diff
@@ -346,32 +486,81 @@ export function getCodecForCategory(category: 'h264' | 'hevc' | 'av1' | 'vp9', w
...
+    for (const profiles of Object.values(CODEC_PROFILES)) {
+        const match = profiles.find(p => p.codec === codec);
+        if (match)
+            return match.name;
     }
 
+    return null;
+}
+
+export function getDefaultCodec(supportedCodecs: CodecInfo[], width: number, height: number): string {
```
alexyakunin: I suspect we need a list here, or at least top 2-3 options, not just 1.

alexyakunin: And the list will be likely somewhat based on device - e.g., iOS devices will definitely prefer VP9 - probably as mobile devices in general.

src/dotnet/Streaming.Service/Backend/LiveVideoBackend.ChatState.cs
```diff
@@ -57,31 +67,76 @@ private void RecomputeSupportedDecoderCodecs(Dictionary<string, ApiArray<string>
             _currentSupportedDecoderCodecs = newCodecs;
         }
 
-        private static ApiArray<string> ComputeSupportedDecoderCodecs(Dictionary<string, ApiArray<string>> members)
+        // Efficiency order, best-first; unknown codecs sort last.
+        private static int CodecPriority(string codec)
```
alexyakunin: Why HEVC goes before vp9 here? Is it used at all?

alexyakunin: Overall, I don't quite understand why server has to do something about codec priority - all it can do it to ensure vp9 is always amount the list of codecs (unless forced).

src/dotnet/Streaming.Service/Backend/LiveVideoBackend.ChatState.cs
```diff
@@ -57,31 +67,76 @@ private void RecomputeSupportedDecoderCodecs(Dictionary<string, ApiArray<string>
...
         {
             if (members.Count == 0)
                 return new ApiArray<string>(["av1", "hevc", "vp9", "h264"]); // No viewers, all codecs available
 
+            // An admin advertising the marker is overriding the negotiation, not
+            // reporting what it can play: its codecs become the call's list as-is,
+            // with no intersection and no floor. Several admins overriding at once
+            // UNION their picks — each is asking for their own codec to be used,
+            // and intersecting would leave nothing.
+            var pinned = members.Values
```
alexyakunin: Maybe call it forced, not pinned?

src/dotnet/Streaming.Service/Backend/LiveVideoBackend.ChatState.cs
```diff
@@ -57,31 +67,76 @@ private void RecomputeSupportedDecoderCodecs(Dictionary<string, ApiArray<string>
...
+            result.Add(FloorCodec); // see FloorCodec
+            if (allSupportH264) result.Add("h264");
+
+            // Members advertise best-first, so order the survivors by the
+            // highest rank any member gave them. Senders take the first entry
+            // they can encode, which is what lets one member pull the call onto
+            // a codec that would otherwise never win: H.264 sits below the
+            // floor on efficiency, so without this it was unreachable no matter
+            // what anyone asked for.
+            var bestRank = new Dictionary<string, int>(StringComparer.Ordinal);
```
alexyakunin: Once more, server fucking shouldn't order this list. The list is UNORDERED by definition, the codec is either supported or not!

PR #4317, opened 2026-08-31
feat(chat): show "typing…" indicators in header, chat list and members

iqmulator:
Ephemeral per-chat typing presence over a Redis-backed channel (mirrors LiveSessions): the editor emits throttled keep-alive pulses, the server tracks typists with a staleness window and self-heals, and the client shows the earliest typist in the chat header, the chat-list preview and the right-panel members list.

Own typing is never shown; a live call/broadcast takes priority over typing in the chat list, and audio recording takes priority in the members list. TypingKind is reserved so the same channel can later carry "sending files…".


alexyakunin:
A few general comments:
- Fuck Redis, it shouldn't be used there. Typing state should be per-chat, in-memory only, so more like online presence state.

src/dotnet/Api.Contracts/Streaming/ITypingIndicators.cs
```diff
@@ -0,0 +1,16 @@
+using ActualChat.Comparison;
+using ActualChat.Live;
+
+namespace ActualChat.Streaming;
+
+public interface ITypingIndicators : IComputeService
```
alexyakunin: Rename to IChatTypingActivities

src/dotnet/Api/Live/TypingKind.cs
```diff
@@ -0,0 +1,9 @@
+namespace ActualChat.Live;
+
+public enum TypingKind
```
alexyakunin: Member 0 should be None, 1 = Typing, and remove unused members.

alexyakunin: rename to TypingActivityKind

src/dotnet/Api.Contracts/Streaming/ITypingIndicators.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    // Authors currently typing in a chat, ordered by when each started their current typing streak -
+    // the UI shows the first one and moves to the next when they stop. Consolidated + compared by
+    // sequence so an unchanged set isn't pushed to every observer.
+    [ComputeMethod(ConsolidationDelay = 0.2, ConsolidationComparer = typeof(ApiArrayComparer<AuthorId>))]
+    [RemoteComputeMethod(CacheMode = RemoteComputedCacheMode.NoCache)]
+    Task<ApiArray<AuthorId>> ListTypingAuthorIds(Session session, ChatId chatId, CancellationToken cancellationToken);
+
+    Task SetTyping(Session session, ChatId chatId, TypingKind kind, bool isActive, CancellationToken cancellationToken);
```
alexyakunin: isActive should be removed (= TypingActivityKind.None)

src/dotnet/Streaming.Service/Backend/TypingIndicatorsBackend.cs
```diff
@@ -0,0 +1,129 @@
...
+        var authorIds = typists
+            .Where(kv => IsFresh(kv.Value, cutoff))
+            .Select(kv => (Ok: AuthorId.TryParse(kv.Key, out var id), Id: id, kv.Value!.StartedAt))
+            .Where(x => x.Ok)
+            .OrderBy(x => x.StartedAt)
+            .Select(x => x.Id)
+            .ToApiArray();
+        if (authorIds.Count > 0)
+            // Re-check so a typist who stopped without an explicit off signal drops on its own.
+            computed.Invalidate(SelfHealDelay);
```
alexyakunin: Once migrated to local, this will be the nearest expiration time.

alexyakunin: And it should be .InvaldiateSafely

src/dotnet/Streaming.Service/Backend/TypingIndicatorsBackend.cs
```diff
@@ -0,0 +1,129 @@
...
+        return authorIds;
+    }
+
+    public virtual async Task SetTyping(
+        ChatId chatId,
+        AuthorId authorId,
+        TypingKind kind,
+        bool isActive,
```
alexyakunin: isActive will go away

src/dotnet/Streaming.Service/Backend/TypingIndicatorsBackend.cs
```diff
@@ -0,0 +1,129 @@
...
+        CancellationToken cancellationToken)
+    {
+        if (isActive) {
+            var now = Clocks.SystemClock.Now;
+            var existing = await SafeGet(chatId, authorId).ConfigureAwait(false);
```
alexyakunin: The storage must be purely local - in fact, it can be protected compute method that simply stores the list, picking up what's primed there, and keeps its own value in RAM due to .Invalidate(...) call, which it always places to make it expire. And coz it's protected, it won't depend on shard owner, etc., i.e. purely local.

src/dotnet/UI.Blazor.App/Services/TypingUI.cs
```diff
@@ -0,0 +1,50 @@
...
+    private ITypingIndicators TypingIndicators => Hub.TypingIndicators;
+    private IAuthors Authors => Hub.Authors;
+    private ConnectivityUI ConnectivityUI => Hub.ConnectivityUI;
+
+    // Authors typing in the chat other than me, ordered by who started first (server order kept).
+    [ComputeMethod]
+    public virtual async Task<ApiArray<AuthorId>> ListTypingAuthorIds(ChatId chatId, CancellationToken cancellationToken)
+    {
+        // While the RPC peer is down we stop receiving invalidations, so the last known value is stale.
+        var isConnected = await ConnectivityUI.IsConnected.Use(cancellationToken).ConfigureAwait(false);
```
alexyakunin: Yep, that's the right thing to do.

src/dotnet/UI.Blazor.App/Services/TypingUI.cs
```diff
@@ -0,0 +1,50 @@
...
+        return ownAuthor is null
+            ? authorIds
+            : authorIds.Where(id => id != ownAuthor.Id).ToApiArray();
+    }
+
+    // The single typist to show - the earliest still typing, or null when nobody is.
+    [ComputeMethod]
+    public virtual async Task<AuthorId?> GetTypingAuthorId(ChatId chatId, CancellationToken cancellationToken)
+    {
+        var authorIds = await ListTypingAuthorIds(chatId, cancellationToken).ConfigureAwait(false);
```
alexyakunin, line 21: If there is >1 author, I'd alternate every 2 seconds or so using even/odd interval semantic (and auto-invalidation here).

src/dotnet/UI.Blazor.App/Services/TypingUI.cs
```diff
@@ -0,0 +1,50 @@
...
+            return authorIds;
+
+        var ownAuthor = await Authors.GetOwn(Session, chatId, cancellationToken).ConfigureAwait(false);
+        return ownAuthor is null
+            ? authorIds
+            : authorIds.Where(id => id != ownAuthor.Id).ToApiArray();
+    }
+
+    // The single typist to show - the earliest still typing, or null when nobody is.
+    [ComputeMethod]
```
alexyakunin: Probably makes sense to consolidate this one w/ 0 delay to avoid excessive renders

src/dotnet/Api.Contracts/Streaming/ITypingIndicators.cs
alexyakunin: Done — `IChatTypingActivities`, and the rest of the set followed: `ITypingIndicatorsBackend` → `IChatTypingActivitiesBackend`, and the impls are now `ChatTypingActivities` / `ChatTypingActivitiesBackend` (files renamed too, so the diff shows them as renames). I left the UI side alone — `TypingUI` and the `ChatTypingIndicator` component — since "typing indicator" is the right word for the widget; say the word if you want those renamed as well.

src/dotnet/Api/Live/TypingKind.cs
alexyakunin: Done: `None = 0`, `Typing = 1`, and `SendingFiles` is gone. It was purely speculative, and with `None` now carrying the "stopped" signal there is nothing to keep it around for.

alexyakunin: Done — `TypingActivityKind`, file renamed to `TypingActivityKind.cs`.

src/dotnet/Api.Contracts/Streaming/ITypingIndicators.cs
alexyakunin: Done. Both contracts are now:

```csharp
Task SetTyping(Session session, ChatId chatId, TypingActivityKind kind, CancellationToken cancellationToken);
```

`TypingActivityKind.None` is the stop signal; the backend removes the author when it sees it. The JS interop shim still passes a bool (`OnTypingChanged(bool isTyping)`) — the editor genuinely only knows on/off — and `ChatMessageEditor` maps it to `Typing`/`None`.

src/dotnet/Streaming.Service/Backend/TypingIndicatorsBackend.cs
alexyakunin: Done — it is the nearest expiration now:

```csharp
var fresh = activities.Where(x => x.ExpiresAt > now).ToApiArray();
if (fresh.Count == 0)
    return default;

computed.InvalidateSafely(fresh.Min(x => x.ExpiresAt) - now, ActivityTtl);
```

The empty case arms nothing: without a `SetTyping` (which invalidates the storage computed and cascades here) no typist can appear out of nowhere.

alexyakunin: Done — `InvalidateSafely(..., ActivityTtl)`, so the delay is clamped and the invalidation carries a source. Same in `ListRaw`.

src/dotnet/Streaming.Service/Backend/TypingIndicatorsBackend.cs
alexyakunin: Gone — the backend signature is now `SetTyping(ChatId, AuthorId, TypingActivityKind, CancellationToken)`, and `kind == None` is what removes the author.

src/dotnet/Streaming.Service/Backend/TypingIndicatorsBackend.cs
alexyakunin: Done — Redis is out entirely, and `RedisMultiHashMap` / `TypingInfo` / the four `Safe*` wrappers went with it. The whole backend is 85 lines now.

The storage is exactly what you described — a protected compute method whose value *is* the list:

```csharp
// Protected, so it never travels over RPC: no shard ownership, nothing to reroute - purely local.
[ComputeMethod]
protected virtual Task<ApiArray<TypingActivity>> ListRaw(ChatId chatId, CancellationToken cancellationToken)
{
    // This is the storage: the value lives in this computed, and the auto-invalidation below both
    // pins it in RAM until the last activity expires and clears it once they all have.
    var computed = Computed.GetCurrent();
    if (!_listRawPrimer.TryUsePrimed(chatId, out var activities) || activities.Count == 0)
        return Task.FromResult(ApiArray<TypingActivity>.Empty);

    computed.InvalidateSafely(activities.Max(x => x.ExpiresAt) - Clocks.SystemClock.Now, ActivityTtl);
    return Task.FromResult(activities);
}
```

`SetTyping` takes the per-chat primer lock, reads the current list through the same method, prunes the expired entries, upserts the author (keeping `StartedAt` across re-emits so the "who started first" order is stable) and primes the result — same `LockingComputeMethodPrimer` shape as `LiveVideoBackend.Register`.

One thing worth flagging, because it's why there are two methods rather than one: the auto-invalidation on the storage method has to fire at the **latest** expiration, not the nearest. `TryUsePrimed` drains the slot, so any recompute that isn't driven by a `Prime` finds nothing and returns empty — if the storage expired at the nearest expiration it would drop the typists that are still fresh along with the one that lapsed. So `ListRaw` holds every entry until they are all dead (that timer is also what pins the value in RAM), and the public `ListTypingAuthorIds` does the freshness filtering and re-checks at the nearest expiration. Both times are bounded by `ActivityTtl` = 6s.

Consequence of going local: a shard handover starts the new owner off empty instead of migrating state. That self-corrects within one keep-alive pulse (~3s), which felt like the right trade for a "typing…" label.

Covered by `tests/Streaming.IntegrationTests/ChatTypingActivitiesTest.cs` — list/stop/order, lapse without a keep-alive, and typing again after a lapse (that last one is the case that would break if the storage computed didn't reset cleanly). All 61 tests in `Streaming.IntegrationTests` pass.

src/dotnet/UI.Blazor.App/Services/TypingUI.cs
alexyakunin: Thanks — kept as is. Without it a dropped RPC peer freezes the last "X is typing…" on screen for as long as the connection is down, since no invalidation can arrive to clear it.

src/dotnet/UI.Blazor.App/Services/TypingUI.cs
alexyakunin: Done — even/odd slots off the absolute clock, with the auto-invalidation landing on the slot boundary:

```csharp
if (authorIds.Count == 1)
    return authorIds[0];

var now = Clocks.SystemClock.Now;
var slot = now.EpochOffset.Ticks / RotationPeriod.Ticks;
var slotEndsAt = new Moment(TimeSpan.FromTicks((slot + 1) * RotationPeriod.Ticks));
computed.Invalidate(slotEndsAt - now);
return authorIds[(int)(slot % authorIds.Count)];
```

`RotationPeriod` is 2s. Deriving the index from the clock rather than from a counter means the invalidation only has to be roughly on time — it can't drift the rotation — and a single typist takes the early return, so a one-person chat arms no timer at all.

I used plain `Invalidate` here rather than `InvalidateSafely`: the delay is bounded by construction (`(0, 2s]`), and `InvalidateSafely`'s 1s floor would make some slots overshoot.

src/dotnet/UI.Blazor.App/Services/TypingUI.cs
alexyakunin: Done — `[ComputeMethod(ConsolidationDelay = 0)]` on `GetTypingAuthorId`. That is the right choke point: the keep-alive pulse re-primes the backend every ~3s, and each pulse used to travel all the way to a re-render even though the author set was identical. Now the recompute happens but the output compares equal and nothing propagates; the only value changes that get through are a real change of typist and the 2s rotation above. Left `IsTyping` at 0.2 as it was.

alexyakunin:
All eleven comments are addressed in `fb6ec72` — replies are inline on each thread.

**Redis is gone.** Typing state is now per-chat, in-memory only, exactly like online presence: the list lives in the value of a protected `ListRaw` compute method, primed through `LockingComputeMethodPrimer` (the same shape `LiveVideoBackend` uses) and pinned in RAM by its own `InvalidateSafely` timer, which is also what expires it. Being protected it never crosses RPC, so it needs no shard-ownership check — the public `ListTypingAuthorIds` keeps that, filters by freshness, and re-checks at the nearest expiration. `RedisMultiHashMap`, the `TypingInfo` record and the four `Safe*` try/catch wrappers all went away with it; the backend is 85 lines.

Also in this commit:
- `ITypingIndicators` → `IChatTypingActivities` (+ backend and both impls, files renamed).
- `TypingKind` → `TypingActivityKind`, `None = 0` / `Typing = 1`, `SendingFiles` removed.
- `bool isActive` dropped from both `SetTyping` contracts — `None` is the stop signal.
- `GetTypingAuthorId` consolidates with 0 delay and rotates between typists every 2s using even/odd clock slots.

New: `tests/Streaming.IntegrationTests/ChatTypingActivitiesTest.cs` (5 tests). The full `Streaming.IntegrationTests` suite is green — 55 passed, 6 skipped. Not exercised in a browser yet, since `/server-loop` isn't mine.

src/dotnet/Streaming.Service/Backend/TypingIndicatorsBackend.cs
alexyakunin: You're right — I had it expiring at the latest expiration precisely so a lapsed author wouldn't take still-fresh ones down with it, and the cost is exactly what you say: one person typing continuously keeps everyone else's dead entries pinned alongside their own. Rebuilt it your way in `fedcde9`.

`ListRaw` no longer self-invalidates. It hands the computed it just produced to a delayed pass armed at the **earliest** expiration:

```csharp
[ComputeMethod]
protected virtual Task<ApiArray<TypingActivity>> ListRaw(ChatId chatId, CancellationToken cancellationToken)
{
    // This is the storage: the value lives in this computed, and the pending ExpireStale below is
    // what keeps it in RAM until then. Nothing primed -> nobody is typing.
    if (!_listRawPrimer.TryUsePrimed(chatId, out var activities) || activities.Count == 0)
        return Task.FromResult(ApiArray<TypingActivity>.Empty);

    var computed = Computed.GetCurrent<ApiArray<TypingActivity>>();
    _ = ExpireStale(chatId, computed, activities.Min(x => x.ExpiresAt));
    return Task.FromResult(activities);
}
```

and the pass itself:

```csharp
await clock.Delay((expiresAt + ExpirationGrace - clock.Now).Positive(), CancellationToken.None)
    .ConfigureAwait(false);
if (!computed.IsConsistent())
    return;

using var isolation = Computed.BeginIsolation();
using var primer = await _listRawPrimer
    .LockAndPrepare(chatId, CancellationToken.None)
    .ConfigureAwait(false);
if (!computed.IsConsistent())
    return;

var activities = computed.Value;
var next = activities.Without(x => x.ExpiresAt <= clock.Now);
if (next.Count != activities.Count)
    await primer.Prime(next, CancellationToken.None).ConfigureAwait(false);
```

Consistent means nothing has replaced the value since, so its survivors are still the truth and it re-primes them; inconsistent means a `SetTyping` (or an earlier pass) already produced a newer list and there is nothing to do. The consistency check is repeated under the primer lock, because a `SetTyping` can land between the delay and acquiring it.

Every update prunes: `SetTyping` already did (`Without(x => x.AuthorId == authorId || x.ExpiresAt <= now)`), and this pass is the other one.

Two consequences worth noting:

- **The pin comes for free now.** The pending `ExpireStale` is the only strong reference the storage computed needs — `ComputedRegistry` holds `WeakReference<Computed>`, so an outstanding task pins it exactly the way the `Invalidate` timer used to, and each re-prime arms the next pass. When the last author lapses the pass primes an empty list, `ListRaw` returns `Empty` without arming anything, and the whole thing falls out of RAM.
- **`ListTypingAuthorIds` dropped its own `InvalidateSafely`.** The nearest expiration is now driven from the storage side and cascades here, so arming a second timer for the same instant would be duplicate work. It keeps the freshness filter, which now only has to cover the `ExpirationGrace` (100ms) window between an expiration and the pass that removes it — the grace is there so the pass can't wake a tick early and find nothing to drop.

No deadlock risk from starting the pass inside a computation that runs under the primer lock: the delay always comes first, and `LockingComputeMethodPrimer` never calls `MarkLockedLocally`, so the re-entry check doesn't apply either. `Computed.BeginIsolation()` keeps the ambient computation that `ListRaw` inherited from leaking dependencies into the pass.

Tests: added `LapsesOneAuthorWhileAnotherKeepsTyping` — one author drops out while the other keep-alives every 2s — for six in `ChatTypingActivitiesTest`. Full `Streaming.IntegrationTests`: 56 passed, 6 skipped. In fairness, the retention this fixes isn't visible through the public API (the freshness filter always hid the stale entries), so that test guards the invalidation cascade rather than the pruning itself.

src/dotnet/Streaming.Service/Backend/ChatTypingActivitiesBackend.cs
```diff
@@ -0,0 +1,116 @@
...
+        // Re-primes the survivors of the earliest expiration, which recomputes ListRaw and arms the
+        // next pass. An inconsistent computed means someone else has already replaced the value.
+        var clock = Clocks.SystemClock;
+        try {
+            await clock.Delay((expiresAt + ExpirationGrace - clock.Now).Positive(), CancellationToken.None)
+                .ConfigureAwait(false);
+            if (!computed.IsConsistent())
+                return;
+
+            using var isolation = Computed.BeginIsolation();
```
alexyakunin: I'd move it before try.

src/dotnet/Streaming.Service/Backend/ChatTypingActivitiesBackend.cs
```diff
@@ -0,0 +1,116 @@
...
+        Computed<ApiArray<TypingActivity>> computed,
+        Moment expiresAt)
+    {
+        // Re-primes the survivors of the earliest expiration, which recomputes ListRaw and arms the
+        // next pass. An inconsistent computed means someone else has already replaced the value.
+        var clock = Clocks.SystemClock;
+        try {
+            await clock.Delay((expiresAt + ExpirationGrace - clock.Now).Positive(), CancellationToken.None)
+                .ConfigureAwait(false);
+            if (!computed.IsConsistent())
```
alexyakunin: This check should go inside LockAndPrepare

src/dotnet/Streaming.Service/Backend/ChatTypingActivitiesBackend.cs
alexyakunin: Done — it is the first statement of the method now, so it covers the delay as well as the primer work:

```csharp
// Re-primes the survivors of the earliest expiration, which recomputes ListRaw and arms the
// next pass. A null primer means someone else has already replaced the value.
using var isolation = Computed.BeginIsolation();
var clock = Clocks.SystemClock;
try {
    ...
```

src/dotnet/Streaming.Service/Backend/ChatTypingActivitiesBackend.cs
alexyakunin: Done, in `ead3124` — I read "inside `LockAndPrepare`" literally and moved *both* checks into the primer rather than just dropping the pre-lock one, since the "take this lock only while the value I'm about to replace is still the current one" rule isn't specific to this backend. `LockingComputeMethodPrimer` gets an additive overload; `LockAndPrepare` itself is untouched, so the existing `LiveVideoBackend` call sites are unaffected:

```csharp
public async ValueTask<Primer?> TryLockAndPrepare(
    TKey key,
    Computed computed,
    CancellationToken cancellationToken = default)
{
    // Returns null once `computed` is inconsistent - its value has been superseded, so there is
    // nothing left to re-prime. Checked before taking the lock, to skip queuing for it at all,
    // and again after, since a writer can land in between.
    if (!computed.IsConsistent())
        return null;

    var primer = await LockAndPrepare(key, cancellationToken).ConfigureAwait(false);
    if (computed.IsConsistent())
        return primer;

    primer.Dispose();
    return null;
}
```

`ExpireStale` now has no consistency checks of its own — the two `if (!computed.IsConsistent()) return;` pairs collapse into one null check:

```csharp
using var primer = await _listRawPrimer
    .TryLockAndPrepare(chatId, computed, CancellationToken.None)
    .ConfigureAwait(false);
if (primer is null)
    return;
```

Two tests in `PrimedComputeServiceTest` cover it: an already-invalidated computed never takes the lock, and one invalidated *while queued* for a lock another holder is sitting on comes back null with no reservation left behind. 23 passing in `Core.Server.UnitTests` Priming, 6 in `ChatTypingActivitiesTest`, solution builds clean.

If you only meant "drop the redundant pre-lock check and keep the one under the lock", say so and I'll revert the shared-infra part.

alexyakunin: Agreed, better — nothing about the gate needs a `Computed`. `94c4c73`:

```csharp
public async ValueTask<Primer?> TryLockAndPrepare(
    TKey key,
    Func<bool> mustProceed,
    CancellationToken cancellationToken = default)
{
    // Returns null once mustProceed stops holding - typically it tracks the consistency of the
    // computed whose value is about to be replaced. Evaluated before taking the lock, to skip
    // queuing for it at all, and again after, since a writer can land in between.
    if (!mustProceed.Invoke())
        return null;

    var primer = await LockAndPrepare(key, cancellationToken).ConfigureAwait(false);
    if (mustProceed.Invoke())
        return primer;

    primer.Dispose();
    return null;
}
```

The consistency case stays a one-liner at the call site, since the extension method converts to `Func<bool>` directly — no lambda:

```csharp
using var primer = await _listRawPrimer
    .TryLockAndPrepare(chatId, computed.IsConsistent, CancellationToken.None)
    .ConfigureAwait(false);
```

Priming tests 23/23, `ChatTypingActivitiesTest` 6/6, solution builds clean.

PR #4331, opened 2026-09-01
feat(chat): add IChats/IChatsBackend.GetTileNonComputed

alexyakunin:
A scan that walks a chat reads a tile per 5 entries. As compute-method calls, each of those becomes a cached, invalidation-tracked slot — and once `IChatsBackend` is a distributed compute service, an RPC-level one that holds an invalidation subscription open. A scan wants the data, not the subscription.

## What this adds

`GetTileNonComputed` as a **plain (non-compute) method on both interfaces** — the same shape as `IUploads.GetOffset`:

```csharp
// IChatsBackend
Task<ChatTile> GetTileNonComputed(
    ChatId chatId, Range<long> lidTileRange, bool includeRemoved, CancellationToken cancellationToken);

// IChats
Task<ChatTile> GetTileNonComputed(
    Session session, ChatId chatId, Range<long> lidTileRange, CancellationToken cancellationToken);
```

Because it is not a `[ComputeMethod]`, a remote call is an ordinary RPC round-trip: no client-side computed, no dependency, no invalidation subscription. The **server implementation still routes through the `GetTile` compute method**, so the tile is read once per node and served from that cache afterwards — the caching is kept, only the tracking is dropped. `Chats.GetTileNonComputed` runs the same `RequireCanRead` check as `GetTile`.

`Computed.BeginIsolation()` in both implementations covers the in-process case, where the caller may itself be a compute method; over RPC there is nothing to isolate.

## Why only the tile

The tile is the read whose key space is large enough for the tracking to matter — one slot per 5 entries, times every scanned chat. `GetIdRange` and the `*RangeMeta` methods are one slot per chat, so tracking them is cheap and they stay compute-only.

## Scope

API surface only. No call site is migrated: on the server, compute methods are usually what you want — every mesh node caches them, so repeated reads stay cheap — and the client-side scan scenarios that would use this aren't identified yet.

## Testing

`ActualChat.CI.slnf` builds clean. New `ChatTileNonComputedTest` (4 tests) pins the semantics: the non-computed call returns the same tile as `GetTile` on both the backend and the session-scoped service; `Computed.Capture` over it captures nothing (with the capturable `GetTile` as the control); and it still throws `NotFoundException` without read permission. Full Chat.IntegrationTests: 402 ✓.

https://claude.ai/code/session_01FMKNBeQWMpoJ8KyCVHzeVZ

### DIFF
--- a/src/dotnet/Chat.Service/Chats.cs
+++ b/src/dotnet/Chat.Service/Chats.cs
@@ -87,6 +87,20 @@ public partial class Chats(IServiceProvider services) : IChats
         return await Backend.GetTile(chatId, lidTileRange, false, cancellationToken).ConfigureAwait(false);
     }
 
+    public async Task<ChatTile> GetTileNonComputed(
+        Session session,
+        ChatId chatId,
+        Range<long> lidTileRange,
+        CancellationToken cancellationToken)
+    {
+        // Isolated so an in-process caller that happens to be a compute method doesn't pick up
+        // the permission check or the tile as dependencies - over RPC there is nothing to isolate.
+        using var _ = Computed.BeginIsolation();
+        await RequireCanRead(session, chatId, cancellationToken).ConfigureAwait(false);
+        return await Backend.GetTileNonComputed(chatId, lidTileRange, false, cancellationToken)
+            .ConfigureAwait(false);
+    }
+
     // [ComputeMethod]
     public virtual async Task<ChatContentSkeleton> GetContentPeriods(
         Session session,
--- a/src/dotnet/Chat.Service/Chats.cs
+++ b/src/dotnet/Chat.Service/Chats.cs
@@ -93,12 +93,11 @@ public partial class Chats(IServiceProvider services) : IChats
         Range<long> lidTileRange,
         CancellationToken cancellationToken)
     {
-        // Isolated so an in-process caller that happens to be a compute method doesn't pick up
-        // the permission check or the tile as dependencies - over RPC there is nothing to isolate.
+        // What this method drops is the caller's dependency, not the caching - hence the isolation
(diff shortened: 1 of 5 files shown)


src/dotnet/Chat.Service/Chats.cs
```diff
@@ -87,6 +87,20 @@ public virtual async Task<ChatTile> GetTile(
...
+        Session session,
+        ChatId chatId,
+        Range<long> lidTileRange,
+        CancellationToken cancellationToken)
+    {
+        // Isolated so an in-process caller that happens to be a compute method doesn't pick up
+        // the permission check or the tile as dependencies - over RPC there is nothing to isolate.
+        using var _ = Computed.BeginIsolation();
+        await RequireCanRead(session, chatId, cancellationToken).ConfigureAwait(false);
+        return await Backend.GetTileNonComputed(chatId, lidTileRange, false, cancellationToken)
```
alexyakunin: Use Backend.GetTile here, we want it to be cached on API server.

alexyakunin: Fixed in b2e0de796e — it now calls `Backend.GetTile` under the isolation.

You're right, and my chaining was wrong on its own terms: what this method drops is the caller's *dependency*, not the caching, so there was no reason to skip the API server's cache of the backend tile. `Computed.BeginIsolation()` already keeps an in-process compute-method caller off that dependency, and the non-compute part that matters is `IChats.GetTileNonComputed` itself being a plain RPC method — that's the client → API server hop where the per-5-entry slots would otherwise pile up.

`IChatsBackend.GetTileNonComputed` is still there for a same-node scan that wants neither the dependency nor (later, cross-mesh) an RPC-level compute slot; it wraps `GetTile`, so it caches on the serving node too.

PR #4356, opened 2026-09-03
Localize system chat titles for Notes, Announcements and onboarding groups

### LINKED ISSUE #4328: Localize chat names for Notes and Announcements

frolyo:
Closes #4328.

## What

System chats are now named in the reader's UI language wherever a chat is titled: the chat list, the header, pickers, share links, push notifications and the email digest.

- **Notes**, the **announcements** chat, a place's **Welcome** chat, and the onboarding groups (**Family, Friends, Classmates / Alumni, Coworkers**).
- Titles stay English in the DB. The frontend `Chats.Get` rewrites `Title` through `ChatTitleLocalizerExt`, the same way it already rewrites a peer chat's title to the peer's name, so every UI surface sees one name. `NotificationsBackend` composes the push title per recipient next to the text, with one lookup per distinct language and no extra reads for ordinary chats.
- **A renamed system chat keeps its name.** `ChatExt.GetSystemDefaultTitle` substitutes only while the stored title is still the default, so an owner who renames Coworkers to "Acme team" sees "Acme team" everywhere, and so does everyone else. Saving the settings form without editing the title does not pin it, because the form diffs against the already-localized chat.
- Onboarding now stores the English default title instead of the current language's label, so the name follows later language changes.
- Each system chat is modelled as `Constants.Chat.SystemChat(Tag, DefaultTitle)` under `Constants.Chat.SystemChats` (formerly `SystemTags`), with an implicit conversion to `Symbol` so tag comparisons read as before.

Catalog: `SystemChat_Notes`, `SystemChat_Announcements_Format` (`{0}` is the product name) and `SystemChat_Welcome` in all 19 hand-written catalogs, bcms and Max regenerated. The onboarding groups reuse their `Onboarding_Chat*` labels.

## Not covered

- Onboarding chats created before this change in a non-English UI hold a localized title in the DB, so they look renamed and stay as they are. A one-off upgrade step could reset those to the English default.
- The search index still holds the stored English title, so searching for a system chat by its localized name does not work yet.

## Tests

- New `SystemChatTitleLocalizationTest` pins the English catalog to the Api defaults, checks every shipped language, a renamed chat and an ordinary chat that merely happens to be called "Notes".
- Localization unit tests (73) and `ChatOperationsTest` integration tests (31) pass. The Notes allowlist entry in `ui-localization-smoke.test.ts` is removed.

### DIFF
--- a/docs/i18n.md
+++ b/docs/i18n.md
@@ -327,6 +327,40 @@ English exists twice on purpose, and `EmptyEntryLocalizationTest` and
 `SystemEntryLocalizationTest` pin the English catalog to each `Default`, so the two
 renderings can't word the same event differently.
 
+### System chat titles
+
+The chats the app creates by itself - the personal Notes chat, the announcements chat,
+a place's Welcome chat, the onboarding groups (Family, Friends, Classmates / Alumni,
+Coworkers) - are stored under the English titles in
+`Constants.Chat.System` (a `SystemChat` is a tag plus its default title) and
+`Constants.Chat.AnnouncementsChatTitle`,
+and named in the reader's language on the way out: the frontend `Chats.Get` rewrites
+`Title` through `ChatTitleLocalizerExt` the same way it rewrites a peer chat's title to
+the peer's name, so the list, the header, pickers and share links all see one name.
+Code with no session localizes per reader - `NotificationsBackend` composes the push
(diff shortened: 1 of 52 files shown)


alexyakunin:
@frolyo merge if it's tested.

PR #4377, opened 2026-09-07
feat(serialization): tolerate unknown [Union] members on read

crui3er:
Adding a member to a `[Union]` hierarchy is currently a breaking change for every peer that doesn't yet know the new tag — and not just for that value: the reader can't know the payload's shape, so it can't even skip it, and the *whole surrounding payload* fails. For a `ChatEntry` that means a tile of a chat rather than one message.

`Markup` has grown to 23 members and `StoredSettings` to 25. These are open sets modelled as closed ones, and every new member has been a compatibility event. This makes an unknown tag survivable, once, for every future member.

## What's here

**`ForwardCompatibleUnionFormatter<TBase>`** reads the tag from a copy of the reader — `MessagePackReader` is a struct over an in-memory sequence, so a copy is a free bookmark. A known tag goes to the standard union formatter untouched, so behaviour on the happy path is bit-for-bit what it was. An unknown one becomes a per-root placeholder, and the envelope is skipped.

Deliberately **not** a try/catch around the inner formatter: that would swallow genuine corruption and turn it into a silently missing message. Corruption with a *known* tag still throws, and there's a test for it.

**Tolerant roots:** `ChatEntry`, `SystemEntry`, `Markup`, `Notification`, `Invite`. `StoredSettings` is deliberately out — tolerance alone doesn't fix its actual bug, which is the unreadable row dropped on write-back, and that's a change to the settings write path. `MuxedAudioStreamItem` and `MediaFrame` are out too: a live stream item has no placeholder that means *less* than a throw, since the demuxer keys off it.

**Per-root policy** is a static abstract interface member (`IForwardCompatibleUnion<TSelf>.NewUnsupported`), so each root's policy lives with the root and `ActualChat.Core` stays free of any API type.

**An unknown `ChatEntry` tag is classified by range** — 100..199 are `SystemEntry` descendants. That matters because `IsSystemEntry` is `this is SystemEntry` and some forty call sites branch on it: a plain `TextEntry` placeholder would take the author-group rendering of a real message and stand as the chat's last message in the list. Both placeholders recover the base prefix — id, version, flags, author — so a tile keeps its keying and an unknown *removed* entry still reads as removed.

**Server-side filtering for peers that can't tolerate anything.** Clients already in the wild don't have the formatter and nothing shipped later can give it to them, so the server stops sending them what they can't read: `GetTile` gains a `LegacyName` twin that drops entries whose kind postdates the peer's API version. A twin rather than a per-peer projection because Fusion's result cache isn't peer-scoped — two methods means two cache slots, and the RPC layer picks which one a peer reaches from the handshake version. Same shape as the `GetNews` / `GetFullNews` pair already in `IChats`.

**Database.** `DbChatEntry.ToModel` no longer throws on a system-entry row whose option it doesn't recognize — a rollback past the writing release used to take out every chat holding one.

## Verified

- `UnionEnvelopeTest` pins the `[tag, payload]` envelope against this codebase's own resolvers rather than assuming it from how MessagePack works. It also turned up that the keyless layout writes derived members *first*, so there is no prefix there and recovery must read by name — anything positional would have silently read a member's field as an id.
- `UnionToleranceCoverageTest` fails any `[Union]` root that is neither tolerant nor exempt with a reason. It immediately found three roots the design had missed.
- `[Union]` attributes survive ILLink trimming: an Android Release publish was linked and its `ActualChat.Api.dll` read out of `obj/.../linked/` — counts identical to the untrimmed assembly. That covers both shipping configurations, since `UseNativeAot` defaults to false and Apple links SDK-only.
- If those attributes are ever trimmed away regardless, `IsTolerant` turns tolerance *off* rather than treating every tag as unknown — the difference between behaving exactly as before this branch and replacing every message in every chat with a placeholder.
- 2 425 unit tests across ten projects, `ActualChat.CI.slnf` builds.

## Before merging

`ApiConstants.LastVersionWithoutUnionTolerance` is `"2.18.9999"`, which assumes this lands in **2.19**. If it slips to 2.20 it must become `"2.19.9999"`, or 2.19 clients get the unfiltered tile.

## Not in scope

`CallOutcomeSystemEntry` is a separate task and becomes the first member that costs nothing beyond its own implementation. Unifying how an entry is stored — `Content` plus a binary `Data` plus a registered discriminator — is its own task too.

### DIFF
--- a/docs/CODING_STYLE.md
+++ b/docs/CODING_STYLE.md
@@ -860,6 +860,17 @@ The key rules:
   properties like `ShardKey`, whose getter can throw on a deserialized instance.
 - **`[Key]` and `[Union]` ordinals are wire format.** Append; never renumber or reuse one on a
   type that has shipped.
+- **Don't reach for `[Union]` for a set that will keep growing.** An unknown enum value costs a
+  reader nothing; an unknown union tag used to be fatal, because the reader can't know the
+  payload's shape and so can't even skip it. `ForwardCompatibleUnionFormatter` makes an unknown
+  tag survivable on the roots that register for it (`UnionToleranceCoverageTest` fails a new root
+  that neither registers nor is exempt with a reason) — but survivable is not *meaningful*: a
+  placeholder is all the reader gets. Where the members are shaped alike, one type with an enum
+  discriminator and a payload keeps the surrounding fields readable, which is strictly better.
+  Reserve `[Union]` for sets whose members genuinely differ in shape.
+- **A `[Union]` member must not reuse a `[Key]` or member name the base declares.** Placeholder
+  recovery reads the base's fields out of a payload it otherwise can't parse; a collision doesn't
+  throw, it silently reads a member's value as an id.
 - **Never add MemoryPack attributes.** MemoryPack no longer writes anything — it only reads
   legacy **stored settings (KVAS)** and **flow state** blobs. A type carrying `[MemoryPackable]`
   either is one of those or is reachable from one; there is no third reason. Anything introduced
--- a/src/dotnet/Api.Contracts/ApiConstants.cs
+++ b/src/dotnet/Api.Contracts/ApiConstants.cs
@@ -8,6 +8,10 @@ public static class ApiConstants
     public static readonly string FullVersionString = typeof(ApiConstants).Assembly.GetInformationalVersion() ?? "0.0+unknown"; // X.Y.Z+123abc
     public static readonly string DisplayVersionString = "v" + FullVersionString.Replace('+', ' ');
     public static readonly Version Version = Version.Parse(VersionString);
+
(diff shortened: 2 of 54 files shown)


crui3er:
Superseded — this landed in `dev` through reworked commits instead of this PR:

- 6c17bd45b5 feat(serialization): tolerate unknown [Union] members on read (reworked version of 3a853adaaf)
- 1aeaf5487c feat(chat): add the CallEntry system entry
- e78f782644 fix(chat): serve a peer that can't read a new entry kind one it can (replaces 98ec69b222: instead of filtering out entry kinds a pre-2.19 peer can't read, it's served a readable substitute)
- 9a0d70e446 docs: forward-compatible unions and call entries

Closing.

PR #4378, opened 2026-09-07
feat(ptt): make gestures reply-only and widen the stop surface

alexis-kochetov:
Makes the PTT gesture model narrower where it opens the mic and wider where it closes it. The asymmetry is the point: a start gesture that misfires transmits your voice, while a stop gesture that misfires just ends your utterance early — and one that *fails* to fire leaves the mic open.

## Start surface: reply-only

Start gestures used to arm for the whole answer window after the app was opened. They no longer do — only voice arms them, so an ordinary jostle can't open a mic just because the app is in the foreground.

Nothing is lost: `AddReplyActions` posts *Reply* on the notification in every non-recording state and it resolves with an unbounded window, so starting a conversation cold still works. This also makes gestures consistent with the headset button, which was already reply-only.

## Your own speech re-arms the window

Without this the window stays dated from the other party's utterance, so a 30s reply of yours consumes most of a 15s window and your follow-up can't be gestured.

`IncomingVoiceActivityUI` becomes `VoiceActivityUI` and gains a parallel own-voice pair, fed by `TrackOwnVoice` watching the edges of `ChatAudioUI.GetRecordingChatId` — the single point every closing path passes through, gesture, record button and notification action alike. `SnapshotLastVoiceAt` merges the two maps and is what every *arming* decision reads; `SnapshotLastIncomingVoiceAt` stays for `PttSession.HandOff`, which replays stamps as incoming and must not inject your own.

Two details worth review attention:

- A reply that heard nothing must not re-arm the triggers that opened it, or one false gesture sustains its own window. `PttReplyUI.CloseReply` calls `SuppressOwnVoiceWindow` **before** `SetRecordingChatId(null)` — the falling edge it suppresses is raised by that very close, so setting the flag afterwards would lose the race.
- `ReplyTargetResolver` also takes the merged snapshot. Once own speech arms the gestures, incoming-only picks the wrong chat in exactly the new case: continue after your own utterance in B, headless so no focused chat, and a stale incoming stamp in A wins.

## Stop surface: added to, not traded

- **Both put-away detections now require proximity coverage.** A phone put down or pocketed covers the sensor; one flipped in mid-air doesn't — that's the false stop this removes. The dwell keeps running while uncovered, so a phone flipped over in the hand fires on landing rather than restarting its 700ms on the surface.
- **`GestureKind.Pocket` split out from `FaceDown`** so the practice panel and the on-device fire logs distinguish them. Detection logic unchanged — the pocket branch stays orientation-agnostic on purpose, since narrowing it to the taught inverted insert would leave the mic open in a pocket entered the other way up.
- **Shake again to stop.** With the mic open a shake can only mean stop, and it was a silent no-op before. It rides the *stop* toggle rather than shake-to-talk — turning off a way to open the mic shouldn't take away a way to close it — and it's gated on `isMicOpen`, not `mustSenseStop`'s wider "transmitting", because a shake sensed for a video-only stream would route to `StartReply` and open the very mic it isn't there to close. In `GestureRecognizer.Process` it runs ahead of the proximity/upside-down guard, like the put-away detector: that guard exists to stop a jostle from opening a mic, not from closing one.
- **Screen lock stops the mic (Android).** `ChatUI.MustKeepAwake` holds `KeepScreenOn` while recording, so the display can't time out under our own activity — which leaves a deliberate power-button press as the reason for an `ACTION_SCREEN_OFF` during a recording, and makes *locking your phone always stops transmission* a one-sentence invariant. Mic only: unlike face-down or pocketing, pressing power says nothing about a camera. `KeepScreenOn` is a window flag, so a foreground *other* app can still time the screen out mid-reply and close the mic early — rare, and the safe direction to fail in. No iOS equivalent; there's no reliable lock event there.

The user-facing model is two gestures over four detections: **put the phone away** (face-down, pocket, screen locked) and **shake again**.

## Rejected

A power-button stop that also fires when the screen was already off: pressing power then means "wake it up so I can look", the opposite of "I'm done", and `ACTION_SCREEN_OFF` wouldn't fire anyway. Gating on a recording (where keep-awake holds) is what removes the ambiguity.

Palm-cover-to-stop: the most natural motion, but it collides with raise-to-ear, since earpiece routing covers proximity. Would need an orientation exclusion that the other two gestures already make unnecessary.

## Testing

- Unit 811/811, UI integration 105 passed / 2 skipped / 0 failed.
- Builds clean: `UI.Blazor.App`, `App.Maui` (net11.0-android), `App.Server`, `App.Wasm`.
- Localization updated across all 19 hand-written catalogs; BCMS and Max regenerated and `--check`-verified.
- `docs/live-audio/10-push-to-talk.md` updated.
- **Verified on device** (CPH2747, `2.19.89-alpha-g74c53c4a7d`): the new gesture behaviour works.

https://claude.ai/code/session_01TyUexScKqeDwY8z4Qp2e7v


alexis-kochetov:
Both review findings were real and are fixed in 36fdb5c488.

**`EndOwnVoice` clearing a valid stamp** — confirmed. The suppression that keeps an unheard reply from re-arming its own triggers was calling `_lastOwnAt.TryRemove`, so a stray gesture landing inside a window your real utterance had opened would end that window early rather than let it run out. An unheard close now neither extends the window nor clears it. The decision moved into a static `ApplyOwnVoiceEnd` so both directions are pinned by tests (`AnUnheardReplyShouldNotCutShortAnEarlierWindow`, `AnUnheardReplyShouldNotExtendTheOwnVoiceWindow`, `AHeardReplyShouldStampItsEnd`).

**Croatian `netko vrijeme`** — confirmed, and it's a `derive-bcms.py` limitation rather than a translation slip. `hr` is generated from `bs`, and the rule list has `("neko ", "netko ")` as a blanket swap; that's right for the pronoun ("netko od vas" later in the same sentence is correct Croatian) but wrong for the determiner in "neko vrijeme". Reworded the Bosnian base to "samo kratko", which sidesteps it — `hr`/`sr`/`cnr` regenerated and `--check`-verified.

Worth noting for later, outside this PR: the script can't express that exception today. Rules are applied longest-source-first, so a whole-phrase entry doesn't shield a substring from a shorter rule — a fix would need a stash/restore pass like the one `to_cyrillic` already uses for `KEEP_LATIN`. Any future string containing "neko vrijeme" will hit the same thing.

Unit tests 814/814.

PR #4472, opened 2026-09-10
feat(attachments): client-side image pipeline with quality presets, placeholders and HEIC

iqmulator:
## Summary

An attached photo was uploaded exactly as the camera wrote it — 4–12 MB per shot, full EXIF including location — and the recipient saw a blank tile until the whole thing arrived and the server had made a thumbnail. HEIC, the default on iPhones and most current Android phones, was worse: Chromium cannot decode it, so it got no preview, no size and went up untouched, and `imageproxy` cannot decode it either, so once stored it was unviewable for every recipient on Chrome or Firefox.

## Fix

**Client-side image pipeline.** Images are resized and re-encoded in a module worker before upload, with a per-draft quality preset: *Up to 12mpx / 6K* (default — a phone's own 12 MP photo passes through unresized), *Up to 3mpx / 3K*, *Original* (metadata stripped) and *Original with EXIF*. The encoder is jpegli compiled to WebAssembly (vendored under `src/nodejs/jpegli`, fetched at runtime, not bundled): ~17% smaller than libjpeg-turbo at equal SSIMULACRA2; a 51 MP encode takes 545 ms on an iPhone 13 Pro, 654 ms on a Galaxy, off the UI thread. Menu sizes come from a fitted model, never a trial encode; the applied row shows the real length once known. Attaching costs nothing — the draft commits on the first of typing, picking a preset, or Send. A GIF is never re-encoded; a source past the server's bounds is stored as a file rather than rejected.

**Placeholders.** Every image carries a 64 px blurred placeholder as bytes on the media row (byte 0 = format mark, byte 1 = signed short side; format 1 rebuilds a fixed 220-byte JPEG prefix so only the entropy-coded payload is stored — median 413 bytes over 15 real photos). `ImagePlaceholder` turns it into the `image-skeleton` data URL, so a recipient sees the picture's shape immediately.

**HEIC.** Where the browser can't decode it (Chromium, Firefox), a vendored libheif WASM (`src/nodejs/libheif`, 469 KB gzipped, fetched only when needed) does; the worker tries `createImageBitmap` first, so WebKit never downloads it. Android decodes natively (`AndroidHeifDecoder`, ImageDecoder → q100 JPEG → the JS pipeline). Two measured subtleties a reviewer must not "fix": EXIF orientation is applied **only** when the file has no `irot`/`imir` (libheif honours the container properties and ignores EXIF; doing both double-rotates every camera photo), and the decode is bounded by the header's dimensions on mobile and declines rather than risking the app. Server-side, `imageproxy` is replaced by imagor (Apache-2.0, libheif built in); nginx rewrites the legacy `/WxH/`, `/xH/`, `/Wx/`, `/0/` URL shapes — `rewrite … last` with a dedicated location and `merge_slashes off`, both of which are load-bearing.

**Platform plumbing.** Android and Apple file providers expose a picked file to `fetch()`; on Windows, WebView2 needs a custom-scheme registration (`TreatAsSecure`, a `Fetch` resource filter, CORS header) before `fetch()` may read `content://`. The MAUI CSP gains `'wasm-unsafe-eval'` — note `index.html` is generated and gitignored; `index.htm` is the source.

**Also fixed along the way:** an upload session leaked its media row when the draft was discarded (now released, but marked bound once a posted entry takes its reference — strictly before the entry is created); Android's gallery picker hands a HEIC over as `application/octet-stream`, so the type is now inferred from the extension; "Choose Photo" / "Select Video" are offered in mobile browsers, not only in the MAUI apps; GIF attachments get dimensions, so their tiles are no longer cropped squares.

**Not implemented:** on-device *video* transcoding. Three rounds of research are distilled into `docs/plans/on-device-video-transcoding.md` (H.264 + AAC passthrough via WebCodecs; AV1/VP9 ruled out) and tracked in #4587.

After the rebase onto the typography overhaul (4773edbcf3) the menu's labels moved onto the semantic type scale (`text-caption` / `text-caption-500`), and the *Up to 50mpx / 12K* preset was dropped from the menu (the enum member, budget and label stay, so a draft that still carries it renders and uploads as before).

## Testing

- TS unit (`vitest`): image-format, image-geometry, image-encoding-policy, metadata-stripper, placeholder-encoder, jpegli-encoder, image-processor-worker, heif-decoder — 73/73 green after the rebase.
- `Chat.UI.Blazor.UnitTests` (Attachment, AttachmentList, ImagePlaceholder, ImageQualityMenuModel, ImageQualityPreset, ImageSizeEstimator, UploadSessions) — 73/73 green.
- `Media.UnitTests` (AttachmentImageUploadProcessor, ImageMetadataStripper, ImageUploadProcessor, MediaKindProcessorSelection, MediaTypeExt, MediaSerialization, MetadataBagJson) — 24/24 green.
- `MediaDbContextModelSnapshot` merged by hand with dev's `AddImageSuggestions`; `dotnet ef migrations has-pending-model-changes` reports none. `AddMediaPlaceholder` (2026-09-13) now sorts before `AddImageSuggestions` (2026-09-16); EF applies it regardless and no deployed DB has it yet.
- Devices: HEIC decode and encode timings measured on Chrome 152 (Windows, Android), iPhone 13 Pro and a Galaxy during development. **Owed:** a fresh end-to-end pass on each app build after the rebase, and a Windows app pass in particular — the WebView2 custom-scheme path has not been exercised end to end. `Media.IntegrationTests` not run locally.


alexyakunin:
@iqmulator FYI I work on this branch - i.e. adding proper resize & encode to JPG.

PR #4494, opened 2026-09-11
fix(threads): notifications, Threads-tab unread badge, mention picker and back button

alexyakunin:
## Summary

These are the four thread bugs from the Bugs chat. The "live calls in threads are hard to find" report is the first one.

1. **No notifications on new thread content.** Only a thread's followers are notified, and the person who started a thread never became one. Since 01fa11a297 threads have no roles. `ContactsBackend.OnChatChangedEvent` still did `RolesBackend.GetSystem(thread, Owner).Require()`, which threw `'Role' is not found` on every thread creation, with several NATS redeliveries each time. The starter's follow was never created. Only the start message's author followed, plus whoever later posted. On top of that, a first @mention of a parent-chat member who didn't follow the thread was dropped. The follow a mention creates comes from a separate event chain, so the notification fan-out still saw a non-follower.
2. **No unread counter on the Threads tab.** The tab was added without a filter, and the badge only rendered for filters. Those counts come from the contact list, which never includes threads.
3. **The mention picker in a thread showed everyone as "not in this chat".** Threads have no authors of their own and no `SeeMembers`, so the member list was empty.
4. **The back button in the thread header** stepped back in history, restoring the parent's old scroll position, or opened the parent at its end.

## Fix

One commit per bug:
- **Notifications.** `ChatThreads.OnStart` creates the starter's thread contact itself, and the dead `ContactsBackend` branch is removed. In `SendChatMessageNotification`, thread mention recipients come from the parent chat's subscribers, filtered to the mentioned users before any per-user mode lookup, so the cost is per mention rather than per member. Plain messages stay followers-only. Author mentions in a thread resolve against its outermost parent, which is what the thread's author ids mirror.
- **Threads tab badge.** `ChatListUI.GetUnreadThreadCount` counts the place's followed threads with the same gated fold the other tabs use. That gate now also covers the selected thread; for plain chats nothing changes. `ThreadListTabUnreadCount` shows the count; clicking it opens the first unread thread.
- **Mention picker.** `LocalSearchUI` resolves a thread to its outermost parent before building candidates, so threads also share the parent's cached lists.
- **Back button.** It always opens `Links.Chat(parent, threadStartLid)`, which scrolls to and highlights the start message.

- **Thread context menus** (follow-up request):
  - **Thread rows** in the left Threads tab and in a chat's Threads panel get a context menu. It has Follow/Unfollow, Mute/Unmute, Go to start message, Copy thread link and Copy thread.
  - **The thread card** in the transcript gains Follow/Unfollow, Mute/Unmute and Copy thread link.
  - **Shared code.** The entries are shared components in `ChatView/ThreadMenu/`.
  - **Mute** is two-state on the thread's own notification settings, the same setting the server's thread notification filter reads.
  - **Localization.** New `ThreadMenu_*` keys are in every catalog; the derived catalogs were regenerated.

Not changed: a call started in a thread still rings only its followers. With the starter now following, participants do get rung. Ringing every parent-chat member would be a product decision, not part of this fix.

## Testing

- New integration test `ChatThreadOperationsTest.StarterShouldFollowThreadStartedOnAnotherUsersMessage`. The whole `ChatThreadOperationsTest` class passes (8/8).
- Live, on a local server with two test users plus a third who was never online:
  - **Before the fix (reproduced).** Starting a thread on another user's message logged the failed `ChatChangedEvent` and left the starter without a follow. The starter then got nothing when the other user posted in the thread.
  - **After the fix: follows and notifications.**
    - The starter follows the new thread, and no queued events failed.
    - Followers get notified of new messages, including the starter.
    - An @mention of a parent member who didn't follow the thread produced a `MentionNotification` for them and made them a follower.
  - **After the fix: UI.**
    - The picker in a thread lists parent members without the "not in this chat" suffix; non-member contacts keep it.
    - The Threads tab shows "1" when a followed thread has unread messages. Clicking it opens that thread, and the badge clears.
    - The back button lands on `?n=<start lid>`, with the start message in view.

Not verified:
- Author-id mentions of anonymous authors in threads at runtime; only user mentions were exercised.
- The highlight animation itself.
- WASM / MAUI; everything ran in server render mode.

https://claude.ai/code/session_01RkXkjXqdQGG32ZNK5K1RqS

### DIFF
--- a/src/dotnet/UI.Blazor.App/Services/ChatListUI.cs
+++ b/src/dotnet/UI.Blazor.App/Services/ChatListUI.cs
@@ -29,6 +29,7 @@ public partial class ChatListUI : UIWorkerBase<AppUIHub>, IComputeService, INoti
     private IContacts Contacts => Hub.Contacts;
     private IAuthors Authors => Hub.Authors;
     private IPlaces Places => Hub.Places;
+    private IChatThreads ChatThreads => Hub.ChatThreads;
     private ActiveChatsUI ActiveChatsUI => Hub.ActiveChatsUI;
     private ChatUI ChatUI => Hub.ChatUI;
     private SearchUI SearchUI => Hub.SearchUI;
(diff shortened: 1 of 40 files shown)


src/dotnet/UI.Blazor.App/Services/ChatListUI.cs
```diff
@@ -31,6 +31,7 @@ private readonly ConcurrentDictionary<Option<PlaceId>, LazySlim<Option<PlaceId>,
     private IPlaces Places => Hub.Places;
     private ActiveChatsUI ActiveChatsUI => Hub.ActiveChatsUI;
     private ChatUI ChatUI => Hub.ChatUI;
+    private IChatThreads ChatThreads => Hub.ChatThreads;
```
alexyakunin: Move it before ActiveChatsUI

alexyakunin: Moved it next to the other API services, before `ActiveChatsUI`, in 562478f168.

PR #4500, opened 2026-09-11
fix(ui): stop Mac trackpad flings from overrunning and jittering at the list end

### LINKED ISSUE #4427: Improve scroll on mac

frolyo:
## Summary

Two Mac Safari trackpad glitches in the chat transcript, both at the newest end of the list (#4427).

**Blank, then a snap to the end.** A fast fling away from the end let the load zone drop the newest item as soon as it was two screens out. With the last item gone the end stopped being known, the bottom limit moved out by three screens, and the fling back ran into blank until the tail reloaded, then snapped to the end in one frame. Engine-independent; reproduced in Chromium with `?vlloaddelay=400`.

**Jitter at the end after a fast return.** The scroll controller claims a precise gesture that starts near an edge and hands it back to the compositor once it is two screens away. On desktop Safari a handed-back gesture cannot be stopped at a limit afterwards: a flick up and a flick back are one gesture (no 200ms gap), so the flick back arrives uncancelable, Safari's inertia runs past the end, and the per-event write back paints a step out and a step home every frame. Measured at 438px worst overshoot, for the length of the inertial tail. Chromium does not do this.

## Fix

- `infinite-list.ts`: a known edge stays in the load zone while it is within `EdgeReachScreens` (12) of the viewport. Six was tried first and one real flick (7.8 screens) dropped the edge again.
- `scroll-controller.ts`: on desktop WebKit (`isWebKit && !isIos`, so Safari and the Mac Catalyst/AppKit WKWebView) every precise gesture is driven from its first event to its last and never handed back (`mustOwnWholeWheelGesture`, override `?vlwheelall=0|1`). Other engines keep the claim-near-edge, hand-back-two-screens-out behaviour. This also removes a one-frame ~216px step that showed at the hand-back moment on Safari.
- `docs/ui/virtual-list.md`: §3.5 and the trackpad section updated with the measurements.

Invariants for reviewers: the fix stays in the list's query builder and in wheel ownership, not in the band/spring path. A rubber band over un-owned trackpad inertia was tried and reverted: on Safari nothing can end the native fling, and it ran 16,073px past the end while the transform tried to absorb it.

## Testing

- `tsc --noEmit`, bundle build and `eslint` on the changed files: green. No unit tests cover these paths.
- Real-finger Safari on macOS, recorder attached to the list: end flag never flips, worst blank 4px, no out/home writes at the end limit; the user confirmed the fling feels right.
- Chromium via the trackpad rig: no blank, no end-flag flip, ends pinned.
- Not verified: Windows precision touchpad in Chromium/Edge (code path unchanged there), and the Mac Catalyst/AppKit app, which shares the WKWebView path and now owns whole gestures too.

Closes #4427

### DIFF
--- a/docs/ui/virtual-list.md
+++ b/docs/ui/virtual-list.md
@@ -1131,6 +1131,19 @@ The claim is therefore made blind and given back later: once a decisive delta sh
 the edge behind — past `WheelOwnScreens` (2) screens from either limit — one event goes uncancelled and
 the compositor scrolls the rest of it, which it does better than the main thread can.
 
+**Desktop WebKit is the exception: every precise gesture is driven whole**, from its first event to its
+last, whatever its distance from a limit (`mustOwnWholeWheelGesture`, `?vlwheelall=0|1`). A gesture
+given back there cannot be stopped at a limit later: its remaining events arrive uncancelable, Safari
+scrolls them off the main thread, and nothing ends the inertia — `killMomentum` is off on WebKit
+(§3.7), and a rubber band over it was measured absorbing 16073px of native travel past the end, a
+disagreement with the compositor on every frame. The case that reaches a limit that way is ordinary on
+a trackpad: a flick up and a flick back are one gesture (no `MotionGapMs` gap), so the flick back
+arrives after the hand-back and runs into the end uncancelable. Snapped per event, that was a step out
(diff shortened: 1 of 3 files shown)


alexyakunin:
@frolyo merged, I made a change to make it render ~ as much as before on non-MacOS (perf).

